PR #4, opened 2020-04-29
Implement Target Flag and List Command

trotttrotttrott:
Implements flag, `-t, --target strings`, in `show`, `diff`, and `apply`. Allows the dashboard targeting by key in a local Jsonnet file.

Implements `grr list`. Lists the dashboard keys of a Jsonnet file.

Closes https://github.com/malcolmholmes/grizzly/issues/2.

### DIFF
--- a/pkg/dash/workflow.go
+++ b/pkg/dash/workflow.go
@@ -127,9 +127,9 @@ local t = %s;
 {
   [k]: { dashboard: f.grafanaDashboards[k], folderId: 0, overwrite: true}
   for k in std.filter(
-		function(n) if std.length(t) > 0 then std.member(t, n) else true,
-		std.objectFields(f.grafanaDashboards)
-	)
+    function(n) if std.length(t) > 0 then std.member(t, n) else true,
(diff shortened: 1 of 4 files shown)


pkg/dash/workflow.go
```diff
@@ -89,15 +100,37 @@ func Apply(config Config, jsonnetFile string) error {
...
+	if err != nil {
+		return nil, err
+	}
+	var keys []string
+	err = json.Unmarshal([]byte(output), &keys)
+	if err != nil {
+		return nil, err
+	}
+	return keys, nil
+}
```
malcolmholmes, line 117: I wonder if the code might be smaller and easier to understand if `dashboardKeys` called renderDashboards, parsed the json and grabbed the keys from that. It would be slower, for sure, but exercises a single code path more and would I suspect be easier to read.

trotttrotttrott: I had considered that, but it assumes that all dashboards will compile. I went this route in case a user is ignoring certain dashboards because they are broken for whatever reason.

PR #5, opened 2020-05-19
Simplify Environment Variables

trotttrotttrott:
Support only `GRAFANA_URL`, `GRAFANA_USER`, `GRAFANA_TOKEN`.

I needed to use an API key and I found the environment variables a bit confusing. Decided to take a pass at simplifying them a little bit.

### DIFF
--- a/pkg/dash/config.go
+++ b/pkg/dash/config.go
@@ -2,7 +2,7 @@ package dash
 
 import (
 	"errors"
-	"fmt"
+	"net/url"
 	"os"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/dash/config.go
```diff
@@ -15,29 +15,22 @@ type Config struct {
...
-		if !hostExists {
-			return nil, errors.New("Either GRAFANA_URL or GRAFANA_HOST required")
-		}
-		if !protocolExists {
-			protocol = "https"
-		}
-		auth := ""
-		if userExists && tokenExists {
-			auth = fmt.Sprintf("%s:%s", user, token)
+	if gu, exists := os.LookupEnv("GRAFANA_URL"); exists {
```
malcolmholmes, line 18:
```suggestion
	if grafanaURL, exists := os.LookupEnv("GRAFANA_URL"); exists {
```
might be a bit more descriptive :-)

pkg/dash/config.go
```diff
@@ -15,29 +15,22 @@ type Config struct {
...
+		u, err := url.Parse(gu)
```
malcolmholmes, line 19: same here, more descriptive variable name please :-)

pkg/dash/config.go
trotttrotttrott: Yeah, "gu" is a little cryptic. Go folks seem to always want to encourage short variable names. For example, https://github.com/golang/go/wiki/CodeReviewComments#variable-names. Being that the declaration and usage are a line apart, I figured I'd make it short as possible.

PR #11, opened 2020-05-26
Prevent panic on diff when remote dashboard is nil

trotttrotttrott:
The `normalize` function would panic when `board.Dashboard` was nil with:

```
panic: assignment to entry in nil map
```

Closes https://github.com/malcolmholmes/grizzly/issues/6


Duologic:
Replaced by #14

PR #13, opened 2020-05-27
feat(watch): apply when file changes

Duologic:
It doesn't work well for watching the same file you want to apply. I'm not 100% sure why. It seem to me that because `vim` removes and writes the file, the remove syscall also removes it from the watch list before the write syscall is captured. But as this project is in an 'alpha' stage, I'd figure this will be alright.


Duologic:
Replaced by #15

PR #14, opened 2020-05-27
Don't panic on missing dashboard

### LINKED ISSUE #6: Diff command panics if remote dashboard not found
I'd like Grizzly to avoid the panic and output a deliberate error message.

malcolmholmes:
closes #6


Duologic:
Looks OK, but when does this happen?

PR #21, opened 2020-06-05
Add export

malcolmholmes:
This adds export, which allows for provisioning

### DIFF
--- a/cmd/grr/workflow.go
+++ b/cmd/grr/workflow.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/go-clix/cli"
 	"github.com/malcolmholmes/grafana-dash/pkg/dash"
 )
@@ -101,6 +103,38 @@ func watchCmd() *cli.Command {
 
 		return dash.Watch(*config, watchDir, jsonnetFile, targets)
 
+	}
+	return cmd
+}
+
+func exportCmd() *cli.Command {
+	cmd := &cli.Command{
+		Use:   "export <jsonnet-file> [-d <dashboard-dir>]",
+		Short: "render Jsonnet and save to a directory",
+	}
+	targets := cmd.Flags().StringSliceP("target", "t", nil, "dashboards to target")
+	dashboardDir := cmd.Flags().StringSliceP("destination", "d", nil, "destination folder")
+	cmd.Run = func(cmd *cli.Command, args []string) error {
+		jsonnetFile := args[0]
+		config, err := dash.ParseEnvironment()
+		if err != nil {
+			return err
+		}
+		var boardDir string
+		boardDirCount := len(*dashboardDir)
+		if boardDirCount == 0 {
+			if config.GrafanaDir != "" {
+				boardDir = config.GrafanaDir
+			} else {
+				return fmt.Errorf("Directory required via -d or GRAFANA_DIR")
+			}
+		} else if boardDirCount == 1 {
+			boardDir = (*dashboardDir)[0]
+		} else if boardDirCount > 1 {
+			return fmt.Errorf("Directory required via -d or GRAFANA_DIR")
+		}
+
+		return dash.Export(*config, jsonnetFile, boardDir, targets)
 	}
 	return cmd
 }
--- a/cmd/grr/workflow.go
+++ b/cmd/grr/workflow.go
@@ -46,12 +46,10 @@ func showCmd() *cli.Command {
 	cmd := &cli.Command{
 		Use:   "show <jsonnet-file>",
 		Short: "render Jsonnet dashboard as json",
+		Args:  cli.ArgsExact(1),
 	}
 	targets := cmd.Flags().StringSliceP("target", "t", nil, "dashboards to target")
 	cmd.Run = func(cmd *cli.Command, args []string) error {
-		if len(args) != 1 {
-			return fmt.Errorf("Expected argument: <jsonnet-file>")
-		}
 		jsonnetFile := args[0]
 		config, err := dash.ParseEnvironment()
 		if err != nil {
@@ -66,12 +64,10 @@ func diffCmd() *cli.Command {
 	cmd := &cli.Command{
 		Use:   "diff <jsonnet-file>",
 		Short: "compare Jsonnet with dashboard(s) in Grafana",
+		Args:  cli.ArgsExact(1),
 	}
 	targets := cmd.Flags().StringSliceP("target", "t", nil, "dashboards to target")
 	cmd.Run = func(cmd *cli.Command, args []string) error {
-		if len(args) != 1 {
-			return fmt.Errorf("Expected argument: <jsonnet-file>")
-		}
 		jsonnetFile := args[0]
 		config, err := dash.ParseEnvironment()
 		if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/grr/workflow.go
```diff
@@ -104,3 +106,35 @@ func watchCmd() *cli.Command {
...
 	return cmd
 }
+
+func exportCmd() *cli.Command {
+	cmd := &cli.Command{
+		Use:   "export <jsonnet-file> [-d <dashboard-dir>]",
+		Short: "render Jsonnet and save to a directory",
+	}
+	targets := cmd.Flags().StringSliceP("target", "t", nil, "dashboards to target")
+	dashboardDir := cmd.Flags().StringSliceP("destination", "d", nil, "destination folder")
```
trotttrotttrott, line 116:
```suggestion
	dashboardDir := cmd.Flags().StringP("destination", "d", "", "destination folder")
```

Seems like we should be anticipating a string instead of []string.

pkg/dash/config.go
```diff
@@ -29,8 +32,6 @@ func ParseEnvironment() (*Config, error) {
 			u.User = url.UserPassword(user, token)
 			config.GrafanaURL = u.String()
 		}
-	} else {
-		return nil, errors.New("Must set GRAFANA_URL environment variable")
```
trotttrotttrott, line 33: Good call here. It'll be nice to skip setting `GRAFANA_URL` for running local-only operations.

cmd/grr/workflow.go
malcolmholmes: Good catch. But removed anyway!

malcolmholmes:
Have removed `-d`. Added some validation. Kept `GRAFANA_DIR` but I see what you mean, it isn't needed. Now add `.json` but then the 'xyz added' is not as right. I wonder if we should set the UID to the key (sans `.json`). That would make life more consistent.

pkg/dash/workflow.go
```diff
@@ -186,6 +189,52 @@ func Watch(config Config, watchDir, jsonnetFile string, targets *[]string) error
...
+		if !strings.HasSuffix(board.Name, ".json") {
+			boardPath += ".json"
+		}
+		existingBoardJSONBytes, err := ioutil.ReadFile(boardPath)
+		existingBoardJSON := string(existingBoardJSONBytes)
+
+		if err != nil && !os.IsNotExist(err) {
+			fmt.Println(err)
+			return err
+		}
```
trotttrotttrott, line 221: Seems like `err` should be checked before type conversion.

pkg/dash/workflow.go
```diff
@@ -186,6 +189,52 @@ func Watch(config Config, watchDir, jsonnetFile string, targets *[]string) error
...
+			fmt.Println(name, green("added"))
+		} else if boardJSON == existingBoardJSON {
+			fmt.Println(name, yellow("unchanged"))
+		} else {
+			fmt.Println(name, green("updated"))
+		}
+		err = ioutil.WriteFile(boardPath, []byte(boardJSON), 0644)
+		if err != nil {
+			return err
+		}
```
trotttrotttrott, line 233: I think `ioutil.WriteFile` should go above the `fmt.Println` messages. If an error occurs, the message will have been incorrect. For example if the output directory doesn't exist:

```
▸ grr export chris.trott/dashboards.jsonnet test
rc-mem-reqs added
2020/06/06 19:35:12 open test/rc-mem-reqs.json: no such file or directory
```

cmd/grr/workflow.go
```diff
@@ -104,3 +124,35 @@ func watchCmd() *cli.Command {
...
+	targets := cmd.Flags().StringSliceP("target", "t", nil, "dashboards to target")
+	cmd.Run = func(cmd *cli.Command, args []string) error {
+		if len(args) == 0 || len(args) > 2 {
+			return fmt.Errorf("Incorrect arguments, expected: <jsonnet-file> [<dashboard-dir>]")
+		}
+		jsonnetFile := args[0]
+		dashboardDir := ""
+		if len(args) == 2 {
+			dashboardDir = args[1]
+		}
```
trotttrotttrott, line 142: go-clix/cli has `ArgsExact` for this! https://github.com/go-clix/cli#positional-arguments...

```suggestion
		Args:  cli.ArgsExact(2),
	}
	targets := cmd.Flags().StringSliceP("target", "t", nil, "dashboards to target")
	cmd.Run = func(cmd *cli.Command, args []string) error {
		jsonnetFile := args[0]
		dashboardDir := args[1]
```

pkg/dash/workflow.go
malcolmholmes: good spot.

pkg/dash/workflow.go
malcolmholmes: yeah, statements are in the past tense, so yes.

pkg/dash/workflow.go
```diff
@@ -186,6 +189,52 @@ func Watch(config Config, watchDir, jsonnetFile string, targets *[]string) error
...
+
+// Export renders Jsonnet dashboards then saves them to a directory
+func Export(config Config, jsonnetFile, dashboardDir string, targets *[]string) error {
+	folderID, err := folderId(config, jsonnetFile)
+	if err != nil {
+		var fID int64 = 0
+		folderID = &fID
+		fmt.Println("Folder not found and/or configured. Applying to \"General\" folder.")
+	}
+	boards, err := renderDashboards(jsonnetFile, targets, *folderID)
```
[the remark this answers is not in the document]
malcolmholmes: Really? We do folders with provisioning - `prometheus-ksonnet` does it. How does it manage it?

pkg/dash/workflow.go
```diff
@@ -186,6 +189,52 @@ func Watch(config Config, watchDir, jsonnetFile string, targets *[]string) error
...
+		if err != nil {
+			return err
+		}
+		boardPath := path.Join(dashboardDir, board.Name)
+		if !strings.HasSuffix(board.Name, ".json") {
+			boardPath += ".json"
+		}
+		existingBoardJSONBytes, err := ioutil.ReadFile(boardPath)
+		if err != nil && !os.IsNotExist(err) {
+			fmt.Println(err)
```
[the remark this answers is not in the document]
malcolmholmes: extraneous debugging, removed.

pkg/dash/workflow.go
```diff
@@ -186,6 +189,52 @@ func Watch(config Config, watchDir, jsonnetFile string, targets *[]string) error
...
+			return err
+		}
+		existingBoardJSON := string(existingBoardJSONBytes)
+
+		err = ioutil.WriteFile(boardPath, []byte(boardJSON), 0644)
+		if err != nil {
+			return err
+		}
+
+		if os.IsNotExist(err) {
```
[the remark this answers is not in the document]
malcolmholmes: you're really paying attention :-)

pkg/dash/workflow.go
```diff
@@ -186,6 +189,52 @@ func Watch(config Config, watchDir, jsonnetFile string, targets *[]string) error
...
+			fmt.Println(name, green("added"))
+		} else if boardJSON == existingBoardJSON {
+			fmt.Println(name, yellow("unchanged"))
+		} else {
+			fmt.Println(name, green("updated"))
+		}
```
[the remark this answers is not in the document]
malcolmholmes: done another way before I saw this...

pkg/dash/workflow.go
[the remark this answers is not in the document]
trotttrotttrott: Check the `grafana-dashboard-provisioning` configmap. That creates our "providers" configurations that maps provisioned dashboards to Grafana folders.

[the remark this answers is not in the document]
malcolmholmes: Okay, so if you also generate the provisioning yaml for dashboards, you can have folders. That feels somewhat out of scope for Grizzly, at least for now.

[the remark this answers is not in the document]
trotttrotttrott: I agree.

PR #23, opened 2020-07-13
Create Dockerfile

yazgoo:


Makefile
```diff
@@ -29,3 +29,6 @@ $(GOX):
 cross: $(GOX)
 	CGO_ENABLED=0 gox -output="dist/{{.Dir}}-{{.OS}}-{{.Arch}}" -ldflags=${LDFLAGS} -arch="amd64 arm64 arm" -os="linux" -osarch="darwin/amd64" ./cmd/g
 
+# Docker container
+container: static
```
trotttrotttrott, line 33: @malcolmholmes we should remove the `static` prerequisite if we're going to also build the `grr` binary as a Docker build step. [Tanka builds the binary with `static`](https://github.com/grafana/tanka/blob/70c2cb39923d7ee738e1cf4eebd7420530809788/Dockerfile#L17-L23) instead with the assumption that the machine running the target has Go installed (which would usually be a CI runner).

Issue #35, opened 2020-08-13
Real hostname


gutorc92:
I deploy Grafana on a kubernetes cluster. I use an ingress to proxy my dns to my container then Grafana hostname stil localhost. On grizzly commands such as preview and apply the url returned is: ```http://localhost:3000/api/snapshots-delete/hzq2FJVuwih1oJTKSk8PsLJyWXqNY2m6 ```.  It will be useful to replace localhost with GRAFANA_HOST env value.

malcolmholmes:
Closing as better options than `preview` (now renamed `snapshot`). Please reopen if still needed.

PR #36, opened 2020-08-13
Real Hostname

gutorc92:
Change hostname returned by snapshot response if it is localhost
to hostname in GRAFANA_URL

Solves #35 
Signed-off-by: Gustavo Coelho <gutorc@hotmail.com>

### DIFF
--- a/pkg/grafana/config.go
+++ b/pkg/grafana/config.go
@@ -7,22 +7,29 @@ import (
 	"path"
 )
 
-func getGrafanaURL(urlPath string) (string, error) {
+func getGrafanaHost() (*url.URL, error) {
 	if grafanaURL, exists := os.LookupEnv("GRAFANA_URL"); exists {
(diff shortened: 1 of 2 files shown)


malcolmholmes:
sorry for my delay. Just checking for localhost doesn't seem right to be. How about `response.hostname != GRAFANA_URL.hostname`? That would make it work for any situations where Grafana is behind a reverse proxy.

gutorc92:
@malcolmholmes, I fixed the code as you suggested.

malcolmholmes:
There are better ways to preview than the `grr preview` command, which needs to be deprecated.

Take a look at the new `grr serve <dir>` (no docs yet though). It'll fire up an http server. Visit it in a browser, you'll be able to preview your resources against a live Grafana instance, but without actually pushing the JSON to that instance.

Closing on the assumption this is no longer relevant. Please feel free to re-open.

PR #37, opened 2020-08-19
feat(*): uids everywhere

sh0rez:
Converts every place a dashboard has to be identified to use the uid.
This is the only reference that is shared between Grafana and Grizzly.

### DIFF
--- a/pkg/grizzly/grafana.go
+++ b/pkg/grizzly/grafana.go
@@ -39,10 +39,7 @@ func (b *Board) UnmarshalJSON(data []byte) error {
 }
 
 func (b Board) UID() string {
-	if s, ok := b.Dashboard["uid"]; ok {
-		return s.(string)
-	}
-	panic("no uid")
+	return b.Dashboard["uid"].(string)
 }
 
 // Boards encasulates a set of dashboards ready for upload
--- a/pkg/grizzly/grafana.go
+++ b/pkg/grizzly/grafana.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strings"
 )
 
 // Folder encapsulates a folder object from the Grafana API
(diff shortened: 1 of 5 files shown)


pkg/grizzly/grafana.go
```diff
@@ -21,17 +22,80 @@ type Folder struct {
 // Board enscapsulates a dashboard for upload to Grafana API
 type Board struct {
 	Dashboard map[string]interface{} `json:"dashboard"`
-	FolderID  int                    `json:"folderId"`
 	Overwrite bool                   `json:"overwrite"`
-	UID       string
-	Name      string
```
malcolmholmes, line 27: don't we want to keep `name`? e.g. for `list`?

pkg/grizzly/grafana.go
```diff
@@ -21,17 +22,80 @@ type Folder struct {
...
+
+	b.Dashboard = m
+	return nil
+}
+
+func (b Board) UID() string {
+	if s, ok := b.Dashboard["uid"]; ok {
+		return s.(string)
+	}
+	panic("no uid")
```
malcolmholmes, line 45: this is probably a bit brief as a message isn't it?
```suggestion
	panic("No UID found. Grizzly requires a uid for all dashboards provided within the Jsonnet")
```

malcolmholmes (approved):

pkg/grizzly/grafana.go
sh0rez: It cannot happen if properly coded. This is only reached in case of a developer error.

Will make more verbose

pkg/grizzly/grafana.go
sh0rez: We maintain that information as the key in the `map[string]Board`. For commands like `list` this is sufficient

PR #38, opened 2020-08-20
feat: Prometheus Rules

sh0rez:
Adds functionality to **parse** Prometheus Rules.

Transitions grizzly away from "dashboards only" to managing multiple (possibly infinite) resource types.
This is reflected in the code by the introduction of the `Resource` interface, a way to treat them all equal in high level code like `Apply`, etc.

Currently, only `Show` and `List` support rules. Other actions will follow in future pull requests.

To keep this small, I decided to file it already. Apologies for the messy nature of this, I'll clean up subsequently.


sh0rez:
CI seems drunk, I can confirm it passes locally and in a fresh docker container

pkg/grizzly/workflow.go
```diff
@@ -34,34 +38,101 @@ func Get(config Config, dashboardUID string) error {
...
+	for _, r := range res {
+		fmt.Fprintf(w, f, r.Kind(), r.UID())
 	}
-	for _, key := range keys {
-		fmt.Println(key)
+
+	return w.Flush()
+}
+
+type Group rulefmt.RuleGroup
```
malcolmholmes, line 57: would it make sense to separate each resource type at least into its own file, if not package?

sh0rez: Agreed. I will do a good cleanup once we have a working state. This will also include package refactoring, similar to how Tanka works

go.mod
```diff
@@ -3,12 +3,13 @@ module github.com/grafana/grizzly
 go 1.13
 
 require (
+	github.com/cortexproject/cortex v1.2.0
```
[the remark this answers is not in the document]
sh0rez: 9MB -> 12MB.

That's still smaller than Tanka and gives us a lot of possibilities. Keep in mind Go only builds into the binary what it needs, not what is in go.mod. This also means this might increase a bit once we start using the Cortex API client

PR #39, opened 2020-08-20
feat(cli): Split-pane pager

sh0rez:
Introduces a custom split-pane pager that allows to select the resource
to be shown on the left side (using mouse our TAB button) and shows the
YAML contents on the right side.

Mouse scroll, Up/Down arrows, HJKL-bindings are supported.

[asciicast](https://asciinema.org/a/YVAgpLGGbU51AQ2e8v3WpvPKX)

This was created out of felt pain of the old way where it printed all resources below each other.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,9 +5,12 @@ go 1.13
 require (
 	github.com/cortexproject/cortex v1.2.0
 	github.com/fatih/color v1.9.0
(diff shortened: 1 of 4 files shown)


Duologic (approved):
Looks good for now to integrate and test it for a while, would like to see it implemented as a `PAGER` or a pr to a great pager like `bat`.

PR #40, opened 2020-08-24
feat(*): Proper diff

sh0rez:
Fixes the diff code to work with the new `Resources` interface and also uses thew new pager when more than 2 resources have differences.

### DIFF
--- a/pkg/grizzly/grafana.go
+++ b/pkg/grizzly/grafana.go
@@ -117,25 +117,6 @@ func (e ErrUidsMissing) Error() string {
 	return fmt.Sprintf("One or more dashboards have no UID set. UIDs are required for Grizzly to operate properly:\n - %s", strings.Join(e, "\n - "))
 }
 
-// // GetAPIJSON returns JSON expected by Grafana API
-// func (b Board) GetAPIJSON() (string, error) {
-// 	b.Overwrite = true
-// 	j, err := json.MarshalIndent(b, "", "  ")
-// 	if err != nil {
-// 		return "", err
-// 	}
-// 	return string(j), nil
-// }
-
-// // GetDashboardJSON returns JSON representation of a dashboard
-// func (b Board) GetDashboardJSON() (string, error) {
-// 	j, err := json.MarshalIndent(b.Dashboard, "", "  ")
-// 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/grizzly/grafana.go
```diff
@@ -105,24 +117,24 @@ func (e ErrUidsMissing) Error() string {
...
+// }
+
+// // GetDashboardJSON returns JSON representation of a dashboard
+// func (b Board) GetDashboardJSON() (string, error) {
+// 	j, err := json.MarshalIndent(b.Dashboard, "", "  ")
+// 	if err != nil {
+// 		return "", err
+// 	}
+// 	return string(j), nil
+// }
```
Duologic, line 137: Is this intentional?

pkg/grizzly/grafana.go
```diff
@@ -22,12 +22,22 @@ type Folder struct {
...
+	if err != nil {
+		panic(err)
+	}
+	return string(data)
 }
 
-func (b Board) String() string {
+type Dashboard map[string]interface{}
+
+func (b Dashboard) String() string {
```
Duologic, line 40: This doesn't seem required for the fix, please break a refactoring apart from the fix so the reviewer can see what is going on.

Duologic (changes requested):
Had a look at the 'fix' commit, couldn't figure out the fix from the refactoring.

pkg/grizzly/grafana.go
sh0rez: No, forgot it. Thanks!

pkg/grizzly/grafana.go
sh0rez: It actually is kind of.

The problem was the fact that the old `Board` type was not able to reflect both, the api representation and the plain one used by Jsonnet.

Unmarshalling both into the same data structure rendered one or the other unusable. Instead of hacking around this, I decided to separate the types, so it works properly.
As a side-effect, diff started to work again

Duologic:
If you can rebase, we can move on with this. Giving that new pager a spin would be interesting.

malcolmholmes:
Closing as this is out of date. We can always bring the features back on top of current code if desired

PR #42, opened 2020-09-02
make grr apply work with folders again

janwillies:
applying dashboards to a folder in grafana didn't work anymore

PR #48, opened 2020-10-07
Fix folders

### LINKED ISSUE #43: grafanaDashboardFolder field does not work
Currently on `v0.0.1`, this field seems to always get ignored. Simplest way to reproduce is adding it to the dashboard in `testdata`, and `grr apply` it:

```
{
  grafanaDashboardFolder: 'FOO',
  grafanaDashboards+:: {
    'my-dash.json': {
      uid: 'prod-overview',
      title: 'Production Overview',
      tags: ['templated'],
      timezone: 'browser',
      schemaVersion: 16,
    },
  },
}
```

The dashboard will be created in _*General*_, and _*FOO*_ will be ignored

malcolmholmes:
Starts from #42 , and pushes it a bit further. This will create a folder if it does not yet exist, and also makes sure that changing a folder results in a re-apply.

Closes #42, closes #43 .

### DIFF
--- a/pkg/grizzly/grafana.go
+++ b/pkg/grizzly/grafana.go
@@ -16,14 +16,15 @@ import (
 
 // Folder encapsulates a folder object from the Grafana API
 type Folder struct {
-	Id    int64
-	Uid   string
-	Title string
+	ID    int64  `json:"id"`
+	UID   string `json:"uid"`
+	Title string `json:"title"`
 }
 
 // Board enscapsulates a dashboard for upload to Grafana API
 type Board struct {
 	Dashboard map[string]interface{} `json:"dashboard"`
+	FolderID  int64                  `json:"folderId"`
 	Overwrite bool                   `json:"overwrite"`
 }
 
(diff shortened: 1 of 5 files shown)


pkg/grizzly/grafana.go
```diff
@@ -124,38 +125,76 @@ func (b Board) GetDashboardJSON() (string, error) {
...
+func (f *Folder) toJSON() (string, error) {
+	j, err := json.MarshalIndent(f, "", "  ")
+	if err != nil {
+		return "", err
+	}
+	return string(j), nil
+}
+
+func getFolder(config Config, UID string) (int64, error) {
+	if UID == "0" {
```
Duologic, line 138: Does this mean the General folder?

pkg/grizzly/grafana.go
```diff
@@ -191,13 +230,18 @@ func getDashboard(config Config, uid string) (*Board, error) {
 
 	type nestedBoard struct {
 		Dashboard Board `json:"dashboard"`
+		Meta      struct {
+			FolderID    int64  `json:"folderId"`
+			FolderTitle string `json:"folderTitle"`
+		} `json:"meta"`
 	}
+
```
Duologic, line 238: Can you drop a comment here on why it is nested? (I think I introduced this but apparently it is less clear as I thought it would be).

Duologic (approved):
LGTM, comments are more for further understanding the code.

pkg/grizzly/grafana.go
malcolmholmes: yep, the default one.

pkg/grizzly/grafana.go
malcolmholmes: that's just how the JSON comes back from the API.

Duologic (approved):

PR #52, opened 2020-10-16
Group resources by handler

malcolmholmes:
Before this PR, the code puts all resources into a single map, regardless of the handler.

With this PR, resources are now grouped by handler. This moves us towards a situation where we hand a whole list of resources to a handler in one go, and the handler is responsible for the whole workflow.

With this change, we're one step closer to being able to implement a `grafanactl` style workflow, independent of jsonnet.

### DIFF
--- a/pkg/grizzly/providers.go
+++ b/pkg/grizzly/providers.go
@@ -45,8 +45,11 @@ func (r *Resource) MatchesTarget(targets []string) bool {
 	return false
 }
 
-// Resources represents a set of resources, by path
-type Resources map[string]Resource
+// ResourceList represents a set of named resources
+type ResourceList map[string]Resource
+
+// Resources represents a set of resources by handler
+type Resources map[Handler]ResourceList
 
 // Handler describes a handler for a single API resource handled by a single provider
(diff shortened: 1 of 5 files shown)


pkg/grizzly/providers.go
```diff
@@ -45,8 +45,11 @@ func (r *Resource) MatchesTarget(targets []string) bool {
 	return false
 }
 
-// Resources represents a set of resources, by path
-type Resources map[string]Resource
+// ResourceList represents a set of named resources
+type ResourceList map[string]Resource
```
Duologic, line 49:
```suggestion
type ResourceMap map[string]Resource
```

Also, `type Resources map[Handler]ResourceList` looks redundant.

Duologic (approved):

pkg/grizzly/providers.go
malcolmholmes: You're right that the naming is rubbish. Naming a map "List" sucks. The structure works well of a map by handler containing a map by name of resources - I just need to work out the naming.

PR #54, opened 2020-10-17
Multi resource handlers

malcolmholmes:
A "multi-resource-handler" is one that accepts all its resources in one go.

Default Grizzly behaviour is to handle resources one at a time, so that the handler does not need to worry about exactly how a diff is evaluated. In most cases, this is fine. However, some handlers require access to all resources in one go. Example cases include:

 * when an API has a batch mode, meaning multiple resources can be delivered to the API server in a single request, making the handler more efficient
 * when a handler processed hierarchical resources, e.g. dashboards and dashboard folders. If the dashboard folder needs to be processed before the dashboard, this allows the handler to deal with this internally. Grizzly just says "here's your resources, do a `diff`, or do an `apply`.

### DIFF
--- a/pkg/grafana/dashboard-handler.go
+++ b/pkg/grafana/dashboard-handler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/grafana/grizzly/pkg/grizzly"
+	"github.com/kylelemons/godebug/diff"
 	"github.com/mitchellh/mapstructure"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/grafana/datasource-handler.go
```diff
@@ -26,29 +26,33 @@ func (h *DatasourceHandler) GetFullName() string {
...
 
-func (h *DatasourceHandler) newDatasourceResource(uid, filename string, source Datasource) grizzly.Resource {
+func (h *DatasourceHandler) newDatasourceResource(path, uid, filename string, source Datasource) grizzly.Resource {
 	resource := grizzly.Resource{
 		UID:      uid,
 		Filename: filename,
 		Handler:  h,
 		Detail:   source,
-		Path:     h.GetJSONPath(),
+		Path:     path,
```
Duologic, line 49: What kind of path is this? Filesystem? Json? URL? Can you make it more succinct?

pkg/grizzly/workflow.go
```diff
@@ -18,6 +18,11 @@ import (
 
 var interactive = terminal.IsTerminal(int(os.Stdout.Fd()))
 
+func isMultiResource(handler Handler) bool {
```
Duologic, line 21: Why not assume it is always a multiResource?

malcolmholmes: because then everything needs to implement Diff/Apply logic. I want to find ways to remove the multi-resource distinction, and make them all do that, but only when I've worked out how to avoid lots of unnecessary code duplication. I have ideas.

pkg/grafana/dashboard-handler.go
```diff
@@ -64,13 +73,84 @@ func (h *DashboardHandler) Parse(i interface{}) (grizzly.ResourceList, error) {
...
+		if err != nil {
+			return err
+		}
+
+		if local == remoteRepresentation {
+			notifier.NoChanges(resource)
+		} else {
+			difference := diff.Diff(remoteRepresentation, local)
+			notifier.HasChanges(resource, difference)
+		}
```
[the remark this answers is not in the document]
malcolmholmes: other handlers have this already. The dashboard-handler is different, in prep for the next PR where dashboards are handled differently, so we can implement folder handling.

pkg/grafana/dashboard-handler.go
```diff
@@ -64,13 +73,84 @@ func (h *DashboardHandler) Parse(i interface{}) (grizzly.ResourceList, error) {
...
+		if resourceRepresentation == existingResourceRepresentation {
+			notifier.NoChanges(resource)
+		} else {
+			err = h.Update(*existingResource, resource)
+			if err != nil {
+				return err
+			}
+			notifier.Updated(resource)
+		}
```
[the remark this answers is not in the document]
malcolmholmes: see above

pkg/grafana/datasource-handler.go
malcolmholmes: yah, fair enough, before this change it was clear it was a jsonPath, but isn't now.

Duologic (approved):

pkg/grizzly/providers.go
```diff
@@ -8,7 +8,7 @@ type Resource struct {
 	Filename string      `json:"filename"`
 	Handler  Handler     `json:"handler"`
 	Detail   interface{} `json:"detail"`
-	Path     string      `json:"path"`
+	JSONPath string      `json:"path"`
```
Duologic, line 11:
```suggestion
	JSONPath string      `json:"json_path"`
```
?

PR #55, opened 2020-10-17
Support old style grafana folders

malcolmholmes:
Now that we have support for multi-resource handlers, we can implement logic to give us back old-style Grafana folder support.

With this, the dashboard handler can receive resources under both `grafanaDashboards` as well as `grafanaDashboardFolder`. Given that the handler is handed all of these resources in one go, before processing the resources, it can check if it has a `grafanaDashboardFolder` element, and if it does, it can set the folder name for all dashboards. In fact, it only does this if the dashboard does not already have a `folderName` set.

This makes Grizzly ready for a release.

### DIFF
--- a/pkg/grafana/dashboard-handler.go
+++ b/pkg/grafana/dashboard-handler.go
@@ -72,7 +72,7 @@ func (h *DashboardHandler) newDashboardFolderResource(path, folderName string) g
 		Filename: folderName,
 		Handler:  h,
 		Detail:   "",
-		Path:     path,
+		JSONPath: path,
 	}
 	return resource
 }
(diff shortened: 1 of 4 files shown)


pkg/grafana/dashboard-handler.go
```diff
@@ -63,9 +66,26 @@ func (h *DashboardHandler) newDashboardResource(path, uid, filename string, boar
...
+		Path:     path,
+	}
+	return resource
+}
+
 // Parse parses an interface{} object into a struct for this resource type
 func (h *DashboardHandler) Parse(path string, i interface{}) (grizzly.ResourceList, error) {
 	resources := grizzly.ResourceList{}
+	if path == dashboardFolderPath {
+		folderName := strings.ReplaceAll(i.(string), "{ }", "") // No idea why json parsing adds { } to the end of the parsed string :-(
```
Duologic, line 84: Might want to have a look at https://github.com/grafana/tanka/pull/358, looks like a similar problem.

Duologic (approved):
This stuff could use some decent unit tests, feels really brittle.

PR #56, opened 2020-10-18
Prometheus rule support

malcolmholmes:
Adds support for the cortex ruler, meaning we can now manage a whole mixin with Grizzly.


pkg/cortex/provider.go
```diff
@@ -0,0 +1,23 @@
...
+type Provider struct{}
+
+// NewProvider returns a new Cortex Provider
+func NewProvider() *Provider {
+	return &Provider{}
+}
+
+// GetName returns the name of the Cortex provider
+func (p *Provider) GetName() string {
+	return "cortex"
```
Duologic, line 15: This looks like a provider to talk to Hosted Prometheus, maybe name it as such?

testdata/cortex-rules.libsonnet
```diff
@@ -0,0 +1,28 @@
+local prom = import 'prom.libsonnet';
+local promRuleGroupSet = prom.v1.ruleGroupSet;
+local promRuleGroup = prom.v1.ruleGroup;
+{
+  prometheus_metamon::
+    promRuleGroup.new('malcolm_alerts')
```
Duologic, line 6:
```suggestion
    promRuleGroup.new('grizzly_alerts')
```

testdata/prom.libsonnet
```diff
@@ -0,0 +1,46 @@
...
+        name: name,
+        rules_map:: {},
+        rules_order:: [],
+        local rules_map = self.rules_map,
+        local rules_order = self.rules_order,
+        rules: [rules_map[rule] for rule in rules_order],
+      },
+
+      rule: {
+        newAlert(name, rule):: {
```
Duologic, line 31:
```suggestion
      newAlertRule(name, rule):: {
```

I don't think this should be nested.

testdata/prom.libsonnet
```diff
@@ -0,0 +1,46 @@
...
+      addGroup(group):: {
+        groups_map+:: {
+          [group.name]: group,
+        },
+        groups_order+:: [group.name],
+      },
+    },
+
+    ruleGroup: {
+      new(name):: {
```
Duologic, line 21:
```suggestion
    newRuleGroup(name):: {
```
Also, no need to next this.

malcolmholmes:
I've renamed this feature from Cortex to Prometheus. These are Prometheus rules. export works with Prometheus. diff/apply/etc works with Cortex as it has an API for uploads.

pkg/cortex/provider.go
malcolmholmes: done, see below.

Duologic (approved):

PR #57, opened 2020-10-25
Fix dashboard preview to ignore folder resources

DeathBorn:
Preview should not be created for dashboard folder resource.


malcolmholmes (approved):
Thanks, makes sense!

DeathBorn:
👌 also, I think I have an idea regarding grafanaFolders problem with `{}`. I think if it would work just like grafanaDashboards, then there will be no problem and on a plus side, additional folder management functionality.

malcolmholmes:
The folders functionality is in place for legacy reasons, the `grafanaDashboardFolder` element is in use.

However, a better design that actually specifies a folder with title and uid, would  be better, especially if it were backwards compatible.

DeathBorn:
Ah.. `grafanaDashboardFolders` would make sense then 🤦

PR #71, opened 2021-01-01
Improve tk show UI

malcolmholmes:
The `tk show` UI, while being clever, has always been frustrating. The key combinations were never clear, particularly.

This PR fixes that. TAB/Shift-TAB, or left/right switch focus between the panes, and up/down will either shift through the resources or scroll the right pane, depending on where the focus is. This now feels much more intuitive and natural to me.

### DIFF
--- a/pkg/term/pager.go
+++ b/pkg/term/pager.go
@@ -5,11 +5,13 @@ import (
 	"github.com/rivo/tview"
 )
 
+// PageItem represents a single item to be viewed
 type PageItem struct {
 	Name    string
 	Content string
 }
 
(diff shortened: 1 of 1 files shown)


Duologic (commented):
I don't see a technical problem with this, but I have no experience with the UI library used here.

malcolmholmes:
> I don't see a technical problem with this, but I have no experience with the UI library used here.

I didn't either! It just adds a few visual cues, and some keyboard shortcuts. A simple test would be to compile it and run `grr show testdata/main.libsonnet` if that makes sense and is useable, then this pr is worthwhile.

PR #77, opened 2021-02-13
Improve handler/provider naming

malcolmholmes:
The objects that underly Grizzly have developed over time. First it just had Providers, then Handlers were added, with Providers representing a group of Handlers.

With all this, we need a standard way of identifying a Handler. Previously, we have used the `jsonPath` for this. But, as we move away from exposing jsonPath to Grizzly, we need a more reliable way to reference both a handler and a resource.

This should be something like `<provider>.<handler>` for a handler, and `<provider>.<handler>.<uid>` for a resource.

This PR:
* gives a Handler awareness of the Provider of which it is part. At present, this is just so that the Handler can report its name in the format listed above.
* makes `GetFullName()` report a compound name as above
* uses `GetFullName()` to implement `resource.Kind()` - a resources' kind is defined by both its provider and its handler.


Duologic:
This sounds a lot like GroupVersionKind in Kubernetes, something we can learn/adopt from there?

malcolmholmes:
> This sounds a lot like GroupVersionKind in Kubernetes, something we can learn/adopt from there?

Good question. It doesn't map to what we have in #72 - as in we have a Prometheus Provider and a Grafana provider that both use `grizzly.grafana.com/v1`.

Perhaps the idea of Providers is overkill at the moment. We had thoughts that Grizzly could cover lots of end-systems, but seem to have settled more on Grafana-related ones (hence Grafana, Cortex Ruler, Synthetic Monitoring).

If we moved the prometheus stuff into the Grafana provider, then everything would sit inside the same provider, and hence we would have a mapping between `apiVersion<->Provider` and `Kind<->Handler`.

malcolmholmes:
@Duologic I followed your idea. This is now fully based on APIVersion (group/version) and Kind, and the Prometheus rules handler is now a part of the Grafana (or should it be Grizzly?) provider. Let me know what you think.

pkg/grafana/rules-handler.go
```diff
@@ -128,3 +132,8 @@ func (h *RuleHandler) Update(existing, resource grizzly.Resource) error {
 	g := resource.Detail.(RuleGroup)
 	return writeRuleGroup(g)
 }
+
+// Preview renders Jsonnet then pushes them to the endpoint if previews are possible
+func (h *RuleHandler) Preview(resource grizzly.Resource, notifier grizzly.Notifier, opts *grizzly.PreviewOpts) error {
+	return grizzly.ErrNotImplemented
```
Duologic, line 138: Is this following golang practices?

Duologic (approved):

pkg/grafana/rules-handler.go
malcolmholmes: hmm, that's a leftover. Will remove in a separate PR.

PR #79, opened 2021-02-18
Use synthetic monitoring api/v1/register/install for authentication

brittenhouse:
Setting to Draft in case this is needed. The better solution is to use the [SM API client](https://github.com/grafana/synthetic-monitoring-api-go-client) for all calls.

Update to use the newer [api/vi/register/install](https://github.com/grafana/synthetic-monitoring-api-go-client/blob/master/docs/API.md#apiv1registerinstall) synthetic monitoring API call for authentication.

See the README for the three additional environment variables that are required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ To interact with Grafana Synthetic Monitoring, you must have these environment v
 
 | Name | Description | Required |
 | --- | --- | --- |
-| `GRAFANA_SM_TOKEN` | Authentication token/api key | true |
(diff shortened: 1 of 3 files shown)


malcolmholmes:
I can't get this to work. If I use an admin api key, on authentication it says I need "metrics publisher" perms. If I use one with metric publisher perms it says I don't have permission to perform this action.

malcolmholmes (approved):
I got this working, and am merging it after a couple of minor tweaks. Thanks!

PR #84, opened 2021-02-28
Set jpath via flag

### LINKED ISSUE #24: Add support of custom Jsonnet library paths
Configuration variable **JsonnetPath** is present in [pkg/grizzly/config.go](https://github.com/grafana/grizzly/blob/master/pkg/grizzly/config.go#L11) but not used. It is really desirable to have the possibility of defining a specific path for Jsonnet libraries.
```golang
// Config provides configuration to `grafana-dash`
type Config struct {
	GrafanaURL  string
	JsonnetPath string
}
```
Also it will be a plus to have the possibility of override the path with the environment variable **JSONNET_PATH**

mavimo:
Closes #24

### DIFF
--- a/cmd/grr/workflow.go
+++ b/cmd/grr/workflow.go
@@ -29,9 +29,11 @@ func listCmd(config grizzly.Config) *cli.Command {
 		Args:  cli.ArgsExact(1),
 	}
 	targets := cmd.Flags().StringSliceP("target", "t", nil, "resources to target")
+	jsonnetPaths := cmd.Flags().StringSliceP("jpath", "J", getDefaultJsonnetFolders(), "Specify an additional library search dir (right-most wins)")
+
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		jsonnetFile := args[0]
-		resources, err := grizzly.Parse(config, jsonnetFile, *targets)
+		resources, err := grizzly.Parse(config, jsonnetFile, *jsonnetPaths, *targets)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


mavimo:
@malcolmholmes ready to be reviewed :)

malcolmholmes:
Thanks for this. This looks (mostly) fine. There's a few things (e.g. whitespace change and `log.SetFlags`) that aren't pertinent to this PR - could we remove them?

My only concern really is that the default jsonnet path values are now cloned to multiple places. Perhaps they could be set within `config` as `DefaultJsonnetPath` and consumed from there in your `--jpath` flag definitions?

mavimo:
@malcolmholmes the reason why is not included in `grizzly.Config` is that this value is set from the command only when is needed (when we process jsonnet file, as happen the file to process), I'm not sure that having this config in the global config is a good idea because:

 - We expose the flag also on command that don't need it (allowing to set config that are not relevant, as should happen in `list` o `providers` commands)
 - We add a property in a config that should change time by time with value that should be nil when not set and we need to understand if in the current context the config should have (or not) the jpath

About log changes you're right, I'm going to remove and open a new PR for that.

mavimo:
@malcolmholmes PR rebased, I would like to squash all commits in a single one, it's fine for you?

malcolmholmes:
Can we somehow define the default jpath in only one place? Other than that, Happy to merge/squash.

mavimo:
@malcolmholmes
> Can we somehow define the default jpath in only one place? 
ok, I'll ping you once done

mavimo:
@malcolmholmes done

cmd/grr/workflow.go
```diff
@@ -210,3 +223,7 @@ func providersCmd(config grizzly.Config) *cli.Command {
 	}
 	return cmd
 }
+
+func getJsonnetFolders() []string {
```
malcolmholmes, line 227: Can we make it this:
```suggestion
func getDefaultJsonnetFolders() []string {
```

Then I think we can merge this.

mavimo: Done :)

mavimo: @malcolmholmes Done :)

mavimo:
@malcolmholmes do you have time to check if something is missing (no pressure, just to know if there is something missing from my side)

malcolmholmes (approved):

Issue #70, opened 2020-12-02
grr apply folder with name not supported


Sanhajio:
`pkg/grafana/dashboards.go|236 func findOrCreateFolder(UID string) (int64, error)` 
only finds folder if the jsonnet file provides UID, 
```
{
  grafanaDashboardFolder: '3d_JKW0Km',
  grafanaDashboards+:: {
+
  },
}
```

while th test data has:
```
{
  grafanaDashboardFolder: 'sample',
  grafanaDashboards+:: {
+
  },
}
```

*Is setting grafanaDashboardFolder by name not supported, or does it need to be implemented ?*

```
> grr --version
grr version v0.0.1-26-g56c3091-dirty
```

haarchri:
in my installation it is working:

`2020/12/17 11:10:04 grr version v0.0.1`

with following syntax:

```
    grafanaDashboardFolder: 'test',
    grafanaDashboards+:: {
        "test.json"+: {
                "uid": "test-001"
        },
```

malcolmholmes:
dashboard folders are a pain. Try a later version (in master - we need to do another release).

malcolmholmes:
@Sanhajio can we close this?

Sanhajio:
@malcolmholmes yes sure please feel free to close it.

PR #85, opened 2021-03-05
Fix folderName injection on dashboard creation

thlacroix:
When a new dashboard was added through `DashboardHandler.Apply`, the folder was not injected in the resource before calling `DashboardHandler.Add`, as this was only done below before calling Update. This PR moves the call to `dashboardWithFolderSet
` before checking if the dashboard already exists, so that it will be done both in the Add and Update cases.

Might be fixing https://github.com/grafana/grizzly/issues/70


malcolmholmes:
The code this PR touches has totally changed. Please check with the latest codebase, and feel free to open another PR if this is still an issue.

PR #86, opened 2021-03-10
Start adding tests

malcolmholmes:
I've got ideas of how Grizzly could grow. But, we have nothing solid in terms of testing to ensure that we don't inadvertently break things for people.

This PR starts with a single (failing!) test. The aim is to establish a working model for creating tests in future.


malcolmholmes:
Further info on how I plan to build out these tests:
* Each Handler will have a test suite that exercises its public functions, e.g. Get, Apply, etc, with a mocked HTTP backend
* The parsing logic will be tested using a mock Provider and mock Handler. When you exercise Diff/Apply/Get etc on this handler it will return responses aimed to exercise the parsing logic/etc.

Duologic (commented):
this looks nice

malcolmholmes:
not going this route

PR #87, opened 2021-03-10
Use tanka native functions

tomwilkie:
Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com>


malcolmholmes (approved):
These are totally necessary. I did add them elsewhere, but clearly they didn't make it into master. Thanks!

malcolmholmes:
Any reason why this is a draft? This code really should be there.

malcolmholmes (approved):

PR #88, opened 2021-03-11
Fix jsonnet unhide for string values

thlacroix:
In the previous version, the jsonnet script to extract private elements didn't work for keys that have string values, like `grafanaDashboardFolder`.

The script did something like:
```
{
  grafanaDashboardFolder:: 'test',
} + {
  grafanaDashboardFolder+::: {}
}
```
and in this case, jsonnet only appends `{ }` as a string to the `grafanaDashboardFolder` value.
Also, all jsonpath from handlers were forced to appear, even if they are not in the source script.

The new script uses a function with the source as argument to check if the field already exists, and force displays it in this case only. The other benefit is that no other handlers keys will be added to the output.

### DIFF
--- a/pkg/grafana/dashboard-handler.go
+++ b/pkg/grafana/dashboard-handler.go
@@ -3,7 +3,6 @@ package grafana
 import (
 	"encoding/json"
 	"fmt"
-	"strings"
 
 	"github.com/grafana/grizzly/pkg/grizzly"
 	"github.com/mitchellh/mapstructure"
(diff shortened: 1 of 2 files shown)


malcolmholmes:
Thanks for this!

The whole idea of extracting hidden elements from Jsonnet is broken - you're just showing one issue amongst many! The right approach is to expose the hidden elements as actual elements from within Jsonnet itself. I've made a library to do this [here](https://github.com/grafana/jsonnet-libs/tree/master/grizzly). And I have a PR that implements this here: #72 . That PR did too many things, and won't be merged. I'm gonna hopefully make another start on it tonight.

In the meantime, it is important we support the old hiddens way, and I'll take a look at this before starting. Thanks again!

Bu

malcolmholmes:
Are you sure this doesn't work? When I run Grizzly with the test data, the dashboard ends up in the `sample` folder. In what way are you seeing this not work?

thlacroix:
Thanks for having a look :)

Before this change, https://github.com/grafana/grizzly/pull/88/files#diff-83b82171360f543df501efc9342cc3475ef71208603a96dba97d096a06598d3dL90 was added to remove the `{ }` suffix, which explains why the folder name appears correct down the line.

However this still mutates the `grafanaDashboardFolder`, so using `$.grafanaDashboardFolder` in dashboards.json would give the result with the wrong suffix.

Also then the `grafanaDashboardFolder` is discarded for dashboard creation, but I've made a separate fix for that https://github.com/grafana/grizzly/pull/85

malcolmholmes:
Grizzly has changed a lot since this PR. Closing, on the assumption this is fixed. Please feel free to re-open.

PR #92, opened 2021-03-13
use constant instead value in logging flag

mavimo:
Use constant for logging to make more clear what our intention in code

PS: there are some specific reason why we want only time and not date & time (default behaviour)?

### DIFF
--- a/cmd/grr/main.go
+++ b/cmd/grr/main.go
@@ -13,7 +13,7 @@ import (
 var Version = "dev"
 
 func main() {
-	log.SetFlags(0)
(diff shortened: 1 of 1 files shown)


malcolmholmes:
> Use constant for logging to make more clear what our intention in code

Makes sense.

> PS: there are some specific reason why we want only time and not date & time (default behaviour)?

No idea. Was inherited from the Tanka project. Mostly I avoid logging for other than for debugging. The one place where it is used is in `watch` (and perhaps `listen` also) as longer running tasks.

malcolmholmes (approved):

PR #93, opened 2021-03-14
update golang version

mavimo:
## Changes
 - Update golang version to 1.16.2 for CI
 - Update golang version to 1.16.2 for docker image

@malcolmholmes I can't sign the `.drone.yaml` can you run `make drone` from this branch? Thanks!

### DIFF
--- a/.drone/drone.jsonnet
+++ b/.drone/drone.jsonnet
@@ -1,4 +1,4 @@
-local golang = 'golang:1.13';
+local golang = 'golang:1.16.2';
 
 local volumes = [{ name: 'gopath', temp: {} }];
 local mounts = [{ name: 'gopath', path: '/go' }];
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -259,4 +259,8 @@ get:
   path: infra/data/ci/docker_hub
   name: password
 
+---
+kind: signature
+hmac: eb0d850b7e3c160dd9ab5f5370f02cccc1c54353834caa55861f30a3ec5d7a0c
+
 ...
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -8,7 +8,7 @@ platform:
 
 steps:
 - name: download
-  image: golang:1.13
+  image: golang:1.16.2
   commands:
   - go mod download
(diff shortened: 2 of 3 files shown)


malcolmholmes (commented):
I'm working on a PR that requires 1.16 right now. Any reason to require 1.16.2 rather than just 1.16?

Also, you need to edit `.drone/drone.jsonnet` then run the `drone` target rather than manually updating `drone.yml`.

I can see if I can sign it. Not sure I've done that before.

mavimo:
@malcolmholmes I prefer to require the specific patch version to make build reproducible. If we use `1.16` different build with the same codebase can produce a different result.

> Also, you need to edit .drone/drone.jsonnet then run the drone target rather than manually updating drone.yml.

is what I do, but since I don't have an account in the drone.grafana.net I can't sign the resulting code.

PS: I dunno how your drone server is configured but drone is able to process jsonnet file so there is no need to dump the yaml file :)

malcolmholmes:
There's various complications regarding Jsonnet in Drone - I don't recall them (e.g. I think to do with importing libraries). Our conclusion is that it is simpler to run the Jsonnet ourselves and commit a YAML.

Can you update the Jsonnet file in this PR?

mavimo:
@malcolmholmes I'm not sure to understand what I need to update, the golang version in the jsonnet file was already updated, see https://github.com/grafana/grizzly/pull/93/files#diff-60045f9474705196e786df20a7bb40d29e013e47a3b6927153a10b9acac82266L1 but when I run `make drone` to dump the YAML file it fail on the signing process (that fail because I'm not in the grafana organization).

I'm missing something?

malcolmholmes:
Okay, I missed that, thanks.

I can sign the file, but I can't add it to this branch. I think you are able to add me to this PR, can you do that? https://stackoverflow.com/questions/44030176/how-to-modify-someone-elses-github-pull-request/45114109

mavimo:
@malcolmholmes "Allow maintainers to edit" was already active, can you double check if it work?

image

malcolmholmes (approved):
Great, thanks for this!

PR #94, opened 2021-03-14
Convert parsing to consume k8s style resources

malcolmholmes:
This is attempt #3. Attempt [#2](https://github.com/grafana/grizzly/compare/k8s-resources-mk2?expand=1) worked beautifully, but was a very large PR to consume.

I managed to split that into two components:
1. Parsing `Resource` objects out of Jsonnet
2. Using Tanka `manifest.Manifest` objects throughout the code.

The latter allows for a lot of code removal, but this can be done as a subsequent PR.

There is one back-compat breakage caused by this PR: elements such as `grafanaDashboards`, etc, *must* now be hidden elements. If they are visible, Grizzly will now report an error (as it will attempt to scan them for k8s style resources, and complain when it fails to find them).


pkg/grafana/datasource-handler.go
```diff
@@ -55,32 +56,32 @@ func (h *DatasourceHandler) newDatasourceResource(path, uid, filename string, so
...
+	source["basicAuthPassword"] = ""
+	source["basicAuthUser"] = ""
+	source["database"] = ""
+	source["orgId"] = 1
+	source["password"] = ""
+	source["secureJsonFields"] = map[string]interface{}{}
+	source["typeLogoUrl"] = ""
+	source["user"] = ""
+	source["withCredentials"] = false
+	source["readOnly"] = false
```
Duologic, line 74: Wouldn't this read cleaner like so:
```
source := Datasource{
  basicAuth: false,
  ...
}
```

pkg/grafana/rules-handler.go
```diff
@@ -56,23 +57,20 @@ func (h *RuleHandler) newRuleGroupingResource(path string, group RuleGroup) griz
...
-// Parse parses an interface{} object into a struct for this resource type
-func (h *RuleHandler) Parse(path string, i interface{}) (grizzly.ResourceList, error) {
+// Parse parses a manifest object into a struct for this resource type
+func (h *RuleHandler) Parse(m manifest.Manifest) (grizzly.ResourceList, error) {
 	resources := grizzly.ResourceList{}
-	msi := i.(map[string]interface{})
-	groupings := map[string]RuleGrouping{}
-	err := mapstructure.Decode(msi, &groupings)
+	spec := m["spec"].(map[string]interface{})
+	group := RuleGroup{}
```
Duologic, line 64:
```suggestion
	group := PrometheusRuleGroup{}
```
Consistency?

pkg/grafana/rules-handler.go
```diff
@@ -21,7 +22,7 @@ func NewRuleHandler(provider Provider) *RuleHandler {
 
 // Kind returns the name for this handler
 func (h *RuleHandler) Kind() string {
-	return "PrometheusRules"
+	return "PrometheusRuleGroup"
```
Duologic, line 25: Why prefix this with `Prometheus` and not Dashboard/Datasource with `Grafana`?

pkg/grafana/synthetic-monitoring-handler.go
```diff
@@ -67,20 +68,19 @@ func (h *SyntheticMonitoringHandler) newCheckResource(path string, filename stri
...
-		check := Check{}
-		err := mapstructure.Decode(v, &check)
-		if err != nil {
-			return nil, err
-		}
-		resource := h.newCheckResource(path, k, check)
-		key := resource.Key()
-		resources[key] = resource
+	spec := m["spec"].(map[string]interface{})
+	check := Check{}
```
Duologic, line 75: Beware that `Check` is a very generic word.

pkg/grizzly/grizzly.jsonnet
```diff
@@ -0,0 +1,86 @@
...
+          'Dashboard',
+          uid(k, dashboards[k]),
+          spec=dashboards[k] + {
+            uid::'',
+          },
+          metadata={ folder: folder }
+        )
+        for k in std.objectFields(dashboards)
+      ];
+      local folder = if 'grafanaDashboardFolder' in main then main.grafanaDashboardFolder else 'General';
```
Duologic, line 30: More newlines please. :innocent:

pkg/grizzly/workflow.go
```diff
@@ -74,61 +76,57 @@ func List(config Config, resources Resources) error {
...
-	return fmt.Sprintf(script, jsonnetFile, strings.Join(handlerStrings, "\n"))
-}
+//go:embed grizzly.jsonnet
+var script string
 
 // Parse evaluates a jsonnet file and parses it into an object tree
 func Parse(config Config, jsonnetFile string, targets []string) (Resources, error) {
 
-	script := getPrivateElementsScript(jsonnetFile, config.Registry.Handlers)
+	script := fmt.Sprintf(script, jsonnetFile)
```
Duologic, line 85: In the Tanka project we noticed the embedded Jsonnet has some awkward side effects: https://github.com/grafana/tanka/issues/522

I'm not sure if this applies here, but if we are going to use top level arguments, we might not want to make the same mistake.

pkg/grafana/datasource-handler.go
malcolmholmes: I keep fixing that one, it never seems to make it in. This is a move rather than a rewrite, but yes, I can fix that.

pkg/grafana/rules-handler.go
malcolmholmes: Because "Dashboard" is obvious whereas "rule" could be anything? Rule is potentially more ambiguous. I suspect that naming is something that could come up in wider discussions.

pkg/grafana/synthetic-monitoring-handler.go
malcolmholmes: Yeah, those types will disappear in future versions, replaced simply by `manifest.Manifest`.

pkg/grizzly/grizzly.jsonnet
malcolmholmes: you're greedy.

pkg/grizzly/workflow.go
malcolmholmes: That's fair, but at present Grizzly doesn't offer any `tla` stuff. I'd expect that, over time, we'll move away from the snippet approach, which is needed because of hidden values anyway (so we already have the problem). This allows us to move towards a place where we don't use them at all (only consuming k8s style resources).

Duologic (approved):

PR #99, opened 2021-03-18
Convert resource type to simple map

malcolmholmes:
Before converting to use kubernetes style objects, the resources Grizzly managed required wrapping, because they needed metadata that isn't present in the resource itself.

Now that we are representing resources kubernetes style, this is no-longer the case. We can now work with a `Resource` object that is just a `map[string]interface{}` (with a few helpers). This allows for the removal of a lot of code making the Grizzly code simpler and easier to read.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /grr
 /dist
 .envrc
+.vscode
--- a/pkg/grafana/synthetic-monitoring-handler.go
+++ b/pkg/grafana/synthetic-monitoring-handler.go
@@ -57,23 +57,23 @@ func (h *SyntheticMonitoringHandler) GetExtension() string {
 func (h *SyntheticMonitoringHandler) Parse(m manifest.Manifest) (grizzly.ResourceList, error) {
 	resource := grizzly.Resource(m)
 	resource.SetSpecString("job", resource.GetMetadata("name"))
-	resource.SetSpecString(folderNameField, resource.GetMetadata("folder"))
+	resource.SetSpecString(folderNameField, resource.GetMetadata("type"))
 	return resource.AsResourceList(), nil
 }
 
 // Unprepare removes unnecessary elements from a remote resource ready for presentation/comparison
 func (h *SyntheticMonitoringHandler) Unprepare(resource grizzly.Resource) *grizzly.Resource {
-	resource = *(resource.DeleteSpecKey("tenantId"))
-	resource = *(resource.DeleteSpecKey("id"))
-	resource = *(resource.DeleteSpecKey("modified"))
-	resource = *(resource.DeleteSpecKey("created"))
+	resource.DeleteSpecKey("tenantId")
+	resource.DeleteSpecKey("id")
+	resource.DeleteSpecKey("modified")
+	resource.DeleteSpecKey("created")
 	return &resource
 }
 
 // Prepare gets a resource ready for dispatch to the remote endpoint
 func (h *SyntheticMonitoringHandler) Prepare(existing, resource grizzly.Resource) *grizzly.Resource {
-	resource = *(resource.SetSpecString("tenantId", existing.GetSpecString("tenantId")))
-	resource = *(resource.SetSpecString("id", existing.GetSpecString("id")))
+	resource.SetSpecString("tenantId", existing.GetSpecString("tenantId"))
+	resource.SetSpecString("id", existing.GetSpecString("id"))
 	return &resource
 }
 
--- a/pkg/grafana/synthetic-monitoring-handler.go
+++ b/pkg/grafana/synthetic-monitoring-handler.go
@@ -1,6 +1,8 @@
 package grafana
 
 import (
+	"fmt"
+
 	"github.com/grafana/grizzly/pkg/grizzly"
 	"github.com/grafana/tanka/pkg/kubernetes/manifest"
 )
@@ -57,7 +59,6 @@ func (h *SyntheticMonitoringHandler) GetExtension() string {
 func (h *SyntheticMonitoringHandler) Parse(m manifest.Manifest) (grizzly.ResourceList, error) {
 	resource := grizzly.Resource(m)
 	resource.SetSpecString("job", resource.GetMetadata("name"))
-	resource.SetSpecString(folderNameField, resource.GetMetadata("type"))
 	return resource.AsResourceList(), nil
 }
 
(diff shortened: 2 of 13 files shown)


pkg/grafana/synthetic-monitoring-handler.go
```diff
@@ -57,96 +53,48 @@ func (h *SyntheticMonitoringHandler) GetExtension() string {
...
 // Unprepare removes unnecessary elements from a remote resource ready for presentation/comparison
 func (h *SyntheticMonitoringHandler) Unprepare(resource grizzly.Resource) *grizzly.Resource {
-	delete(resource.Detail.(Check), "tenantId")
-	delete(resource.Detail.(Check), "id")
-	delete(resource.Detail.(Check), "modified")
-	delete(resource.Detail.(Check), "created")
+	resource = *(resource.DeleteSpecKey("tenantId"))
+	resource = *(resource.DeleteSpecKey("id"))
+	resource = *(resource.DeleteSpecKey("modified"))
+	resource = *(resource.DeleteSpecKey("created"))
```
Duologic, line 69: This looks like we could have a different representation of resource.

```
return resource.asUnprepared()
```

pkg/grafana/synthetic-monitoring-handler.go
```diff
@@ -57,96 +53,48 @@ func (h *SyntheticMonitoringHandler) GetExtension() string {
...
 	return &resource
 }
 
 // Prepare gets a resource ready for dispatch to the remote endpoint
 func (h *SyntheticMonitoringHandler) Prepare(existing, resource grizzly.Resource) *grizzly.Resource {
-	resource.Detail.(Check)["tenantId"] = existing.Detail.(Check)["tenantId"]
-	resource.Detail.(Check)["id"] = existing.Detail.(Check)["id"]
+	resource = *(resource.SetSpecString("tenantId", existing.GetSpecString("tenantId")))
+	resource = *(resource.SetSpecString("id", existing.GetSpecString("id")))
```
Duologic, line 76:
```
return resource.asPrepared()
```

pkg/grafana/synthetic-monitoring.go
```diff
@@ -71,22 +71,24 @@ func getRemoteCheck(uid string) (*Check, error) {
 		return nil, err
 	}
 	for _, check := range checks {
-		if check.UID() == uid {
+		if check.Job() == uid {
```
Duologic, line 74: In some places I see UID replace by `Name` and here it is by `Job`, is that intended?

pkg/grafana/synthetic-monitoring.go
```diff
@@ -183,24 +185,24 @@ func getProbeList() (*Probes, error) {
...
-}
-
-// UID retrieves the UID from a check
-func (c *Check) UID() string {
+// Job retrieves the job name from a check
+func (c *Check) Job() string {
 	job, ok := (*c)["job"]
 	if !ok {
-		return "X"
+		return ""
```
Duologic, line 192: Can you add a comment why it would be OK to return an empty string?

pkg/grizzly/providers.go
```diff
@@ -6,40 +6,102 @@ import (
...
+	y, err := yaml.Marshal(*r)
+	if err != nil {
+		return "", err
+	}
+	return string(y), nil
+
+}
+
+// GetRepresentation Gets the string representation for this resource
+func (r *Resource) GetRepresentation() (string, error) {
```
Duologic, line 99: I still find this a confusing function name.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /grr
 /dist
 .envrc
+.vscode
```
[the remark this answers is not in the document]
malcolmholmes: Clever (and obvious)

pkg/grafana/synthetic-monitoring-handler.go
malcolmholmes: interesting, yes. But, it would still need to call out to a func within the handler to actually do it.

pkg/grafana/synthetic-monitoring.go
malcolmholmes: Yeah, this is an internal thing to SM, I suspect this code could be simplified though, potentially loosing the `Check` struct.

pkg/grafana/synthetic-monitoring.go
malcolmholmes: It absolutely isn't. But these happen in places where functions don't return errors. That does need resolving, no solution yet.

pkg/grizzly/providers.go
```diff
@@ -6,40 +6,102 @@ import (
...
+func (r *Resource) GetMetadata(key string) string {
+	metadata := (*r)["metadata"].(map[string]interface{})
+	return metadata[key].(string)
+}
+
+func (r *Resource) SetMetadata(key, value string) Resource {
+	metadata := (*r)["metadata"].(map[string]interface{})
+	metadata[key] = value
+	(*r)["metadata"] = metadata
+	return *r
```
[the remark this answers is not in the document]
malcolmholmes: That's fair. Yes.

pkg/grizzly/providers.go
```diff
@@ -6,40 +6,102 @@ import (
...
-func (r *Resource) GetRemoteRepresentation() (string, error) {
-	return r.Handler.GetRemoteRepresentation(r.UID)
+func (r *Resource) AsResourceList() ResourceList {
+	key := r.Key()
+	resources := ResourceList{}
+	resources[key] = *r
+	return resources
+}
+
+func (r *Resource) SpecAsJSON() (string, error) {
```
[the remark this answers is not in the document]
malcolmholmes: No, because this just returns the spec element, as JSON, not the whole thing. `toJSON` could be fine, but what about `toYAML`, does that return the spec, or the whole resource?

pkg/grizzly/providers.go
malcolmholmes: It can probably be replaced by `YAML()` now actually.

pkg/grizzly/providers.go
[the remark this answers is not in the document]
Duologic: If Resource.Spec is a type, you could setup `func (s Spec) toJSON() (string,error)`

Duologic (approved):

PR #100, opened 2021-03-19
Add YAML file and directory support

malcolmholmes:
Up until now, Grizzly has been all about Jsonnet. You code your resources
in Jsonnet (or if you must, JSON), and you proceed from there.

This blocks a number of use-cases. Firstly, it requires a niche language.
Secondly, that niche language has to be rendered before it is useful.
Being able to simply parse files as YAML as well as Jsonnet/JSON would
open up a range of more static possibilities.

This PR adds YAML support. If a filename ends with `.yaml` or `.yml`, it
will be parsed as, well, YAML.

Beyond that, we don't necessarily want to define all of our resources in
a single file. In fact, we may have a LOT, and may want to split them
over multiple directories.

This PR also adds support for reading directories. If, in its parsing, it
comes across a Directory resource, it will use it to search for suitable
files that it will also parse. An example YAML could be:

```
kind: Directory
metadata:
  name: dashboards
spec:
  path: resources/dashboards
```

Alternatively, you can specify a glob:
```
kind: Directory
metadata:
  name: dashboards
spec:
  glob: resources/dashboards/*
```
Files identified in this way can be either YAML or Jsonnet/JSON. They are parsed the same way as any other file.

Given this, it is important to watch out for using a `Directory` resource that references itself, as this will cause an endless loop.


Duologic (commented):
Related to PR #109

PR #109, opened 2021-04-03
Add support for running Grizzly with a config file

malcolmholmes:
Adds support for a basic config file to drive Grizzly. This will facilitate a number of features to come yet.

An example config file could be:
```
kind: GrizzlyConfig
metadata:
  name: grizzly
spec:
  outbound:
  - name: tests
    path: testdata/yaml/*.yaml
```
Executing Grizzly like this:
```
grr show -c grizzly.yaml
```
would cause Grizzly to consume this yaml file, and expand the 'outbound' globs, then parse each file in turn. At present, this mode only supports YAML. For Jsonnet support, just carry on as before. Jsonnet doesn't make sense in this context, because of its `import` functionality.

This PR has been carefully coded to be reviewed one commit at a time. Each commit should be a distinct, and easily consumed change.

### DIFF
--- a/cmd/grr/main.go
+++ b/cmd/grr/main.go
@@ -26,22 +26,18 @@ func main() {
 		log.Fatalln(err)
 	}
 
-	config := grizzly.Config{
-		Registry: registry,
-		Notifier: grizzly.Notifier{},
-	}
 	// workflow commands
 	rootCmd.AddCommand(
-		getCmd(config),
-		listCmd(config),
-		showCmd(config),
-		diffCmd(config),
-		applyCmd(config),
-		watchCmd(config),
-		listenCmd(config),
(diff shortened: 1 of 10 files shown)


Duologic (commented):
As discussed in person, I think you can prevent the use of 'globs' here. 

I understand the problem is that we want to write a number of (yaml) objects by a 'pull' mechanism and then read those objects (yaml files here) with a 'push' mechanism. Instead of assuming the glob matches all files the 'pull' mechanism has produced, I would propose that the 'pull' mechanism has a way to communicate that list of files to the 'push' mechanism, the most straight forward example I can come up with is updating a static manifest file with a list of these objects (yaml files).

This removes the current need for a fixed config file.

PR #111, opened 2021-04-11
Prepare for pull functionality

malcolmholmes:
This PR tidies up a number of things that either make the code easier
to read/work with, or resolves issues found using Grizzly against a
large Grafana instance.

It is best reviewed one commit at a time:

- Support YAML parsing
- Switch to GrizzlyOpts
- Reduce code duplication
- Move parsing into separate file
- Remove unnecessary embedding
- Replace Config with just Registry
- Actually remove the Config struct
- Fix args(!) and prep for config file
- Rename, tidy
- add missing error check
- Support getting datasources by UID
- Treat datasource ID as an integer
- Again, ID as int not string
- Handle folders with spaces in name
- Improve dashboard folder uid generation
- Use UID for datasources first, then defer to name

### DIFF
--- a/pkg/grafana/datasources.go
+++ b/pkg/grafana/datasources.go
@@ -13,7 +13,7 @@ import (
 
 // getRemoteDatasource retrieves a datasource object from Grafana
 func getRemoteDatasource(uid string) (*grizzly.Resource, error) {
-	grafanaURL, err := getGrafanaURL("api/datasources/name/" + uid)
+	grafanaURL, err := getGrafanaURL("api/datasources/uid/" + uid)
 	if err != nil {
 		return nil, err
 	}
@@ -25,7 +25,7 @@ func getRemoteDatasource(uid string) (*grizzly.Resource, error) {
 	defer resp.Body.Close()
 	if resp.StatusCode == http.StatusNotFound {
 		resp.Body.Close()
-		grafanaURL, err := getGrafanaURL("api/datasources/uid/" + uid)
+		grafanaURL, err := getGrafanaURL("api/datasources/name/" + uid)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/grafana/datasources.go
+++ b/pkg/grafana/datasources.go
@@ -11,9 +11,8 @@ import (
 	"github.com/grafana/grizzly/pkg/grizzly"
 )
 
-// getRemoteDatasource retrieves a datasource object from Grafana
-func getRemoteDatasource(uid string) (*grizzly.Resource, error) {
-	grafanaURL, err := getGrafanaURL("api/datasources/uid/" + uid)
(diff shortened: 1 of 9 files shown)


pkg/grafana/datasources.go
```diff
@@ -23,7 +23,19 @@ func getRemoteDatasource(uid string) (*grizzly.Resource, error) {
...
+		grafanaURL, err := getGrafanaURL("api/datasources/name/" + uid)
+		if err != nil {
+			return nil, err
+		}
 
+		resp, err = http.Get(grafanaURL)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
```
Duologic, line 37: I'd DRY this code a bit.

pkg/grizzly/config.go
```diff
@@ -1,9 +1,17 @@
...
+	ConfigFile   *string
+	ResourceFile *string
+	JsonnetPaths *[]string
+	Targets      *[]string
+}
+
+func (o *GrizzlyOpts) ConsumeArguments(args []string) {
+	if len(args) > 0 {
+		o.ResourceFile = &args[0]
+	}
```
Duologic, line 14: I wouldn't abstract this into a separate function.

pkg/grizzly/parsing.go
```diff
@@ -0,0 +1,106 @@
...
+	"strings"
+
+	"github.com/google/go-jsonnet"
+	"github.com/grafana/tanka/pkg/jsonnet/native"
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
+	"github.com/grafana/tanka/pkg/process"
+	"gopkg.in/yaml.v3"
+)
+
+func Parse(registry Registry, opts GrizzlyOpts) (Resources, error) {
```
Duologic, line 20: I'd propose building an interface that implements a parser, similar to [the 'Loader' interface in Tanka](https://github.com/grafana/tanka/blob/master/pkg/tanka/load.go#L117). It uses a 'Detect` function to figure out which Parser to execute. All logic seems to exist here already, just have to restructure.

```go
type Parser interface {
	Parse(registry Registry, file string, opts GrizzlyOpts) (Resources, error)
}
```

pkg/grizzly/providers.go
```diff
@@ -177,35 +177,37 @@ type Provider interface {
...
 	if !exists {
 		return nil, fmt.Errorf("No handler registered to %s", path)
 	}
 	return handler, nil
 }
+
+// Notifier returns a notifier for responding to users
+func (r *Registry) Notifier() *Notifier {
+	return &Notifier{}
+}
```
Duologic, line 213: Bit weird, almost feels like Java. :stuck_out_tongue:

Duologic (approved):

Duologic (approved):
:heart:

PR #112, opened 2021-04-11
Add directory parsing and Pull support

malcolmholmes:
Adds support for locating resources within a fixed directory structure.

Usage:
```
$ grr show -d testdata/yaml
```

The files are expected to follow this pattern:
```
dashboards/$FOLDER_ID/dashboard-$UID.yaml
datasources/datasource-$UID.yaml
prometheus/rules-$UID.yaml
synthetic-monitoring/check-$UID.yaml
```

Also adds 'pull' support. `grr pull -d resources` will pull any resources
found within the remote systems into a `resources/` directory, using
the same structure as mentioned above.

- Add directory parsing support
- Use Stringer rather than Resource for notifications
- More concise Stringify
- Add pull support
- Add YAML test data


cmd/grr/workflow.go
```diff
@@ -24,9 +24,9 @@ func getCmd(registry grizzly.Registry) *cli.Command {
 
 func listCmd(registry grizzly.Registry) *cli.Command {
 	cmd := &cli.Command{
-		Use:   "list <jsonnet-file>",
+		Use:   "list [-d <directory> [<jsonnet-file>]",
```
sh0rez, line 27: Nit: The file vs dir logic might become simpler if the `-d` flag was of type bool instead, so that the commands would always accept a single arg and the flag controls how it is interpreted

pkg/grafana/dashboards.go
```diff
@@ -52,6 +52,49 @@ func getRemoteDashboard(uid string) (*grizzly.Resource, error) {
...
+		defer resp.Body.Close()
+
+		switch resp.StatusCode {
+		case http.StatusNotFound:
+			return nil, grizzly.ErrNotFound
+		default:
+			if resp.StatusCode >= 400 {
+				return nil, errors.New(resp.Status)
+			}
+		}
```
sh0rez, line 78:
```suggestion
		switch resp.StatusCode {
		case http.StatusNotFound:
			return nil, grizzly.ErrNotFound
		case resp.StatusCode >= 400:
			return nil, errors.New(resp.Status)
		}
```

pkg/grafana/dashboards.go
```diff
@@ -52,6 +52,49 @@ func getRemoteDashboard(uid string) (*grizzly.Resource, error) {
...
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+		var dashboards []Dashboard
+		if err := json.Unmarshal([]byte(string(body)), &dashboards); err != nil {
+			return nil, err
+		}
```
sh0rez, line 86: `ioutil.ReadAll` buffers in memory, which is unnecessary for your use here. You have an `io.Reader`, no need for buffering:

```suggestion
		var dashboards []Dashboard
		if err := json.NewDecoder(resp.Body).Decode(&dashboards); err != nil {
			return nil, err
		}
```

pkg/grafana/datasources.go
```diff
@@ -56,6 +56,42 @@ func getRemoteDatasource(uid string) (*grizzly.Resource, error) {
 	return &resource, nil
 }
 
+func getRemoteDatasourceList() ([]string, error) {
```
sh0rez, line 59: Same comments as above, no `if` in `default`, use `json.NewEncoder`

pkg/grizzly/parsing.go
```diff
@@ -18,13 +20,50 @@ import (
...
-		strings.HasSuffix(*opts.ResourceFile, ".libsonnet") ||
-		strings.HasSuffix(*opts.ResourceFile, ".json") {
-		return ParseJsonnet(registry, *opts.ResourceFile, opts)
+	if opts.ResourceFile != nil && opts.Directory != nil {
+		return nil, fmt.Errorf("Only specify a directory or a resource file, not both")
+	}
+	if opts.ResourceFile != nil {
+		return ParseFile(registry, opts, *opts.ResourceFile)
+	}
+	if opts.Directory != nil {
```
sh0rez, line 29: If we've made it this far, `opts.Directory` is always `!= nil`, you can drop that if

pkg/grizzly/parsing.go
```diff
@@ -18,13 +20,50 @@ import (
...
+		resources := Resources{}
```
sh0rez, line 30: Super Nit: no need for a struct literal if you don't wanna set values
```suggestion
		var resources Resources
```

pkg/grizzly/parsing.go
```diff
@@ -18,13 +20,50 @@ import (
...
+			}
+			resources = append(resources, r...)
+		}
+		return resources, nil
+	}
+	return nil, fmt.Errorf("Must specify either a resource file or a directory")
+}
+
+func FindResourceFiles(registry Registry, opts GrizzlyOpts, dir string) ([]string, error) {
+	files := []string{}
```
sh0rez, line 48:
```suggestion
	var files []string
```

pkg/grizzly/parsing.go
```diff
@@ -18,13 +20,50 @@ import (
...
+	if strings.HasSuffix(resourceFile, ".yaml") ||
+		strings.HasSuffix(resourceFile, ".yml") {
+		return ParseYAML(registry, resourceFile, opts)
+	} else if strings.HasSuffix(resourceFile, ".jsonnet") ||
+		strings.HasSuffix(resourceFile, ".libsonnet") ||
+		strings.HasSuffix(resourceFile, ".json") {
+		return ParseJsonnet(registry, resourceFile, opts)
 	} else {
 		return nil, fmt.Errorf("Either a config file or a resource file is required")
 	}
```
sh0rez, line 69: This screams for a switch! Also there is `filepath.Ext` ;)

```suggestion
func ParseFile(registry Registry, opts GrizzlyOpts, resourceFile string) (Resources, error) {
	switch filepath.Ext(resourceFile) {
	case ".yml", ".yaml":
		return ParseYAML(registry, resourceFile, opts)
	case ".jsonnet":
		return ParseJsonnet(registry, resourceFile, opts)
	}
	
	return nil, fmt.Errorf("Either a config file or a resource file is required")
```

sh0rez (approved):

cmd/grr/workflow.go
malcolmholmes: yeah, I thought about that recently. We can add `-H` or `--hidden` which switches on hidden element parsing in jsonnet, and remove it from the current Jsonnet approach. Switching to bools helps here.

pkg/grafana/dashboards.go
malcolmholmes: really? My IDE tells me that's not valid (`cannot convert resp.StatusCode >= 400 (untyped bool value) to int`)

sh0rez: oh my bad, sorry for that. The following should do:

```go
switch {
    case resp.StatusCode == http.StatusNotFound:
      // sth
    case resp.StatusCode >= 400:
      // sth
}

File: pkg/grizzly/config.go [hunk 9]
```diff
@@ -2,8 +2,8 @@ package grizzly
 
 // GrizzlyOpts contains options for all Grizzly commands
 type GrizzlyOpts struct {
-	Directory    *string
-	ResourceFile *string
+	ResourcePath *string
```
sh0rez, line 5: ResourcePath appears mandatory (it can't work without, right?) It shouldn't be in `opts` then, let's leverage the fact that function arguments are mandatory as well and compile-time checked

pkg/grizzly/config.go
```diff
@@ -2,8 +2,8 @@ package grizzly
...
+	Directory    *bool
 	JsonnetPaths *[]string
 	Targets      *[]string
```
sh0rez, line 8: Why are these pointers? slices are pointers anyways, so you double-pointer here, and the bool doesn't need to be one as well, right?

Also consider you risk panicking when using those variables without checking they are non nil. No need to use unsafe constructs here ;)

sh0rez (changes requested):

cmd/grr/workflow.go
```diff
@@ -24,16 +25,14 @@ func getCmd(registry grizzly.Registry) *cli.Command {
...
```
sh0rez, line 28:
```suggestion
		Use:   "list <path>",
```

These are neither optional (`[...]`), nor does `-d` take an argument ;)

cmd/grr/workflow.go
```diff
@@ -24,16 +25,14 @@ func getCmd(registry grizzly.Registry) *cli.Command {
...
 		Short: "list resource keys from file",
 		Args:  cli.ArgsExact(1),
 	}
-	opts := grizzlyOptsFromCmd(cmd)
+	opts := grizzly.GrizzlyOpts{}
```
sh0rez, line 32:
```suggestion
	var opts grizzly.GrizzlyOpts
```

This form is very common for "just initializing". Also, why isn't that `grizzly.Opts` which reads nicer?

cmd/grr/workflow.go
```diff
@@ -218,12 +221,10 @@ func providersCmd(registry grizzly.Registry) *cli.Command {
 	return cmd
 }
 
-func grizzlyOptsFromCmd(cmd *cli.Command) grizzly.GrizzlyOpts {
-	return grizzly.GrizzlyOpts{
-		ConfigFile:   cmd.Flags().StringP("config", "c", "", "config file"),
-		Targets:      cmd.Flags().StringSliceP("target", "t", nil, "resources to target"),
-		JsonnetPaths: cmd.Flags().StringSliceP("jpath", "J", getDefaultJsonnetFolders(), "Specify an additional library search dir (right-most wins)"),
-	}
+func defaultGrizzlyFlags(opts *grizzly.GrizzlyOpts, fs *pflag.FlagSet) {
```
sh0rez, line 224: What about:

```go
func defaultGrizzlyFlags(fs *pflag.FlagSet) grizzly.GrizzlyOpts {
    var opts grizzly.GrizzlyOpts
    // ...
    return opts
}
```

go.mod
```diff
@@ -18,6 +18,7 @@ require (
 	github.com/mitchellh/mapstructure v1.3.3
 	github.com/pmezard/go-difflib v1.0.0
 	github.com/rivo/tview v0.0.0-20200818120338-53d50e499bf9
+	github.com/spf13/pflag v1.0.5 // indirect
```
sh0rez, line 21: Run `go mod tidy` please, this dependency is not really indirect

sh0rez (approved):

pkg/grizzly/workflow.go
```diff
@@ -70,6 +72,48 @@ func List(registry Registry, resources Resources) error {
 	return w.Flush()
 }
 
+// Pulls remote resources
+func Pull(registry Registry, opts GrizzlyOpts) error {
+
+	f, _ := os.Create("log")
+	defer f.Close()
```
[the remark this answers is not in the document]
malcolmholmes: removed

cmd/grr/workflow.go
malcolmholmes: sure, adjusted.

cmd/grr/workflow.go
malcolmholmes: Yeah, I reworked all of those, and the descriptions which still mention jsonnet

cmd/grr/workflow.go
malcolmholmes: Only if you can explain why that didn't work when I tried it.

PR #116, opened 2021-04-30
Add 10 second timeout to all HTTP clients

julienduchesne:
Otherwise, calls can hang forever

### DIFF
--- a/pkg/grafana/httpclient.go
+++ b/pkg/grafana/httpclient.go
@@ -1,0 +1,10 @@
+package grafana
+
+import (
+	"net/http"
+	"time"
+)
+
+func NewHttpClient() *http.Client {
+	return &http.Client{Timeout: 10 * time.Second}
+}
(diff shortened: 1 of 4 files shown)


pkg/grafana/httpclient.go
```diff
@@ -0,0 +1,10 @@
+package grafana
+
+import (
+	"net/http"
+	"time"
+)
+
+func NewHttpClient() *http.Client {
+	return &http.Client{Timeout: 10 * time.Second}
```
dannykopping, line 9: Would be even nicer if this were configurable

sh0rez: I second this, might render grizzly unusable in bad network conditions otherwise

malcolmholmes: `GRIZZLY_HTTP_TIMEOUT` with a default of 10s?

dannykopping: Seems reasonable to me

julienduchesne: Done

dannykopping (approved):
Nice one, thanks @julienduchesne :+1:

malcolmholmes (approved):

PR #117, opened 2021-05-05
Get grafanabot token from Vault instead of Drone repository

julienduchesne:
https://github.com/grafana/deployment_tools/issues/9828


jvrplmlmn (approved):
👍

PR #120, opened 2021-05-17
Add ability to list (and filter) remote resources

malcolmholmes:
`grr list` currently tells you the resources available locally within your Jsonnet or YAML.

This PR adds the ability to list remote resources, in the same format. This was surprisingly easy to add, all the infrastructure (e.g. pulling a list of resources) was already present.

### DIFF
--- a/pkg/grizzly/providers.go
+++ b/pkg/grizzly/providers.go
@@ -220,7 +220,7 @@ func (r *Registry) GetHandler(path string) (Handler, error) {
 }
 
 // HandlerMatchesTarget identifies whether a handler is in a target list
-func (r *Registry) HandlerMatchesTarget(handler Handler, targets []string, isSilent bool) bool {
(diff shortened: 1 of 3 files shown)


pkg/grizzly/providers.go
```diff
@@ -231,7 +231,9 @@ func (r *Registry) HandlerMatchesTarget(handler Handler, targets []string) bool
 			return true
 		}
 	}
-	r.Notifier().Info(SimpleString(key), "skipped")
+	if !isSilent {
```
Duologic, line 234: Where is this false? And shouldn't this log message move there then instead?

malcolmholmes: yeah, that's actually fair. Notifying shouldn't be a side-effect of an "Is" function.

Duologic (approved):

PR #122, opened 2021-06-09
Add support for watching directories recursively

### LINKED ISSUE #121: Add ability to watch directories recursively
Many mixins use non-flat directory structure (e.g. dashboards are housed in a separate directory), which makes watching problematic. There are workarounds (running multiple processes, or touching a file in a watched directory), but they are inconvenient.

It looks like `fsnotify` does not yet support userspace recursive watcher (https://github.com/fsnotify/fsnotify/issues/18), but as an interim solution the watcher can walk the directory and set up separate watchers on each subdir.

The proposed solution will make watching less efficient, but my assumption is that it won't hog the performance too much.

radiohead:
Closes #121 

This PR adds ability to watch directories recursively, by using `filepath.WalkDir` to list subdirectories and adding them to fsnotify Watcher.

The functionality is hidden behind a flag, which is wired using custom options, similar to how preview command options are implemented.

The functionality is disabled by default.

Signed-off-by: Igor Suleymanov <igor.suleymanov@grafana.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,12 @@ $ grr apply my-lib.libsonnet
 
 ### grr watch
 Watches a directory for changes. When changes are identified, the
-jsonnet is executed and changes are pushed to remote systems. This
-example watches the current directory for changes, then executes
+jsonnet is executed and changes are pushed to remote systems.
+The directory is watched recursively (i.e. all subdirectories are watched too),
+but if new subdirectories are added, watch command needs to be re-started,
+as new directories will not be picked up automatically.
+
+This example watches the current directory for changes, then executes
 `my-lib.libsonnet` when changes are noticed:
 
 ```sh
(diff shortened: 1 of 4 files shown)


malcolmholmes:
This is great! My only question is whether we need the recursive flag. Would people expect it to work recursively already, in which case this is a big fix rather than a feature?

radiohead:
> This is great! My only question is whether we need the recursive flag. Would people expect it to work recursively already, in which case this is a big fix rather than a feature?

I'm one to suggestions. For this PR I've decided to err on the side of caution and make it fully backwards-compatible, but if you think we should simply go ahead and make recursive watching the default, I'm happy to update it.

malcolmholmes:
I think make it the default, and note in the docs that it will watch existing dies but not spot new ones without a restart?

radiohead:
> I think make it the default, and note in the docs that it will watch existing dies but not spot new ones without a restart?

Done!

malcolmholmes (approved):

malcolmholmes:
Thanks for your contributions!

PR #123, opened 2021-06-10
Fix apply errors when using folder name with spaces

radiohead:
This PR fixes an issue that happens when a mixin is using a dashboard folder name which contains spaces, e.g.:

```jsonnet
{
  grafanaDashboardFolder:: 'My Awesome Folder',

  grafanaDashboards+:: {
    'some-dashboard': (import './dashboards/some-dashboard.libsonnet'),
  },
}
```

The root cause is the fact that spaces are removed from the name when preparing the Folder resource in Jsonnet, but the name is used as-is when preparing Dashboard resources.

I took the liberty of fixing what looks to be a bug in `FolderHandler.GetByUID` and improving error messages and error handling in the folder handler.

### DIFF
--- a/pkg/grizzly/workflow.go
+++ b/pkg/grizzly/workflow.go
@@ -120,7 +120,7 @@ func Pull(registry Registry, resourcePath string, opts Opts) error {
 				continue
 			}
 			resource, err := handler.GetByUID(UID)
-			if errors.As(err, &ErrNotFound) {
+			if errors.Is(err, ErrNotFound) {
 				registry.Notifier().NotFound(SimpleString(UID))
 				return nil
 			}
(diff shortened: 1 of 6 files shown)


pkg/grizzly/workflow.go
```diff
@@ -220,40 +225,48 @@ func Apply(registry Registry, resources Resources) error {
 	for _, resource := range resources {
 		handler, err := registry.GetHandler(resource.Kind())
 		if err != nil {
+			// Why?
```
malcolmholmes, line 228: Fair question. We should return a "handler not found for this Kind" type error.

malcolmholmes (approved):
All totally reasonable changes. Wanna fix the `// why?? ` while you're at it?

pkg/grizzly/workflow.go
radiohead: I'll do that, thanks for the suggestion. Also, I totally did not intend to commit these 😆

radiohead: Done.

malcolmholmes (approved):
Looks great, thanks!

PR #124, opened 2021-06-14
Fixup 'grr get' examples in README.md

rhowe:
Since (I think) introducing Kubernetes-style resource naming, the examples in
README.md no longer work as advertised.

`grr get dashboard.foo` is now `grr get Dashboard.foo`


malcolmholmes (approved):
Thanks for this. Meant to fix for ages.

PR #126, opened 2021-06-25
Handle general folder better

malcolmholmes:
The "General" folder is special. It has an ID of zero, but, actually doesn't exist in Grafana. You can't access the General folder via the HTTP API. Here, we make Grizzly "fake" the general folder.

### DIFF
--- a/pkg/grafana/folders.go
+++ b/pkg/grafana/folders.go
@@ -13,6 +13,16 @@ import (
 
 // getRemoteFolder retrieves a folder object from Grafana
 func getRemoteFolder(uid string) (*grizzly.Resource, error) {
+	h := FolderHandler{}
+	if uid == "General" || uid == "general" {
+		folder := Folder{
+			"id":    0.0,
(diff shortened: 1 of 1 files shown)


pkg/grafana/folders.go
```diff
@@ -13,6 +13,16 @@ import (
 
 // getRemoteFolder retrieves a folder object from Grafana
 func getRemoteFolder(uid string) (*grizzly.Resource, error) {
+	h := FolderHandler{}
+	if uid == "General" || uid == "general" {
+		folder := Folder{
+			"id":    0.0,
```
dsotirakis, line 19: Is there a reason for the id to be float?

dsotirakis (approved):
LGTM, one question though

pkg/grafana/folders.go
malcolmholmes: simply because that's what the JSON lib expects.

PR #128, opened 2021-06-30
Get folder UID from Dashboard Metadata

dsotirakis:
This PR is responsible for extracting the folder UID from `Dashboard.Meta` struct.

If the UID isn't present, then it's extracted from the folder url.


pkg/grafana/folders.go
```diff
@@ -159,6 +159,39 @@ func putFolder(resource grizzly.Resource) error {
...
+
+	resp, err := http.Get(grafanaURL)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case http.StatusNotFound:
+		return nil, fmt.Errorf("NOT FOUND")
```
malcolmholmes, line 176: I think we have an `ErrNotFound` we can use. I recognise that this isn't your doing thought - I'm sure this is wrong in other places too.

testdata/yaml/prometheus/rules-simple.yaml
```diff
@@ -1,16 +0,0 @@
...
-    rules:
-        - alert: PromScrapeFailed
-          annotations:
-            message: Prometheus failed to scrape a target {{ $labels.job }}  / {{ $labels.instance }}
-          expr: up != 1
-          for: 1m
-          labels:
-            severity: critical
-        - expr: sum by(job) (up)
-          record: job:up:sum
```
malcolmholmes, line 16: we wanna keep these files :-)

testdata/yaml/synthetic-monitoring/check-simple.yaml
```diff
@@ -1,44 +0,0 @@
...
-        - Sydney
-    settings:
-        http:
-            failIfNotSSL: false
-            failIfSSL: false
-            ipVersion: V4
-            method: GET
-            noFollowRedirects: false
-    target: https://grafana.com/
-    timeout: 2500
```
malcolmholmes, line 44: same for this one :-)

malcolmholmes (approved):
LGTM

PR #130, opened 2021-07-01
Add test suites

dsotirakis:
The scope of this PR is to add test suites for every untested piece of code. This includes unit tests and tests against `dashboards.go`, `folders.go` etc., which need to mock http clients respectfully.

The tests are meant to be run against a Grafana instance (with custom config and provisioned resources), which will run as a service on drone, and all integration tests will run against it.


pkg/grafana/run_container.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+docker run --name grafana \
+           -v $(pwd)/testdata:/etc/grafana \
+           -e GF_PATHS_CONFIG=/etc/grafana/custom.ini \
+           -e GF_PATHS_PROVISIONING=/etc/grafana/provisioning \
+           --rm -p3000:3000 \
+           grafana/grafana:8.0.1
```
malcolmholmes, line 8: might want to go up to 8.0.4 now?

pkg/grafana/testdata/custom.ini
```diff
@@ -0,0 +1,10 @@
+[auth.anonymous]
+enabled = true
+org_role = Admin
+
+[feature_toggles]
+# enable features, separated by spaces
+enable = live
```
malcolmholmes, line 7: do we need this anymore?

malcolmholmes (approved):
This looks amazing. My only thought is that we should move the current `testdata` into an `examples` dir, and then your stuff makes sense in the `testdata` dir.

PR #132, opened 2021-07-01
[synthetic monitoring] replace some api call

ying-jeanne:
it is not optimized, for keeping the handler as it is i am not initialize the client once and reuse it, will do a second pr for that.


dsotirakis:
+1,366 lines in go.sum 😱

ying-jeanne:
this is actually not backward compatible, since the field probes is used for probeid, then the old probes become probenames. I don't think it worths to be make it backward compatible.

pkg/grizzly/providers.go
```diff
@@ -92,6 +98,21 @@ func (r *Resource) Spec() map[string]interface{} {
...
+	}
+
+	err = json.Unmarshal(data, &smSpec)
+	if err != nil {
+		return synthetic_monitoring.Check{}, nil
+	}
+
+	return smSpec.Check, nil
+}
+
```
malcolmholmes, line 115: does this even work? Really, this functionality needs to be in `grafana/synthetic-monitoring.go`, rather than having handler specific stuff in the generic package.

testdata/yaml/synthetic-monitoring/check-simple.yaml
```diff
@@ -11,7 +11,7 @@ spec:
     job: grafana-com
     labels: []
     offset: 0
-    probes:
+    probeNames:
```
malcolmholmes, line 14: this is a breaking change. A reasonable one, but breaking. Can we make it back-compatible?

ying-jeanne: done

pkg/grizzly/workflow.go
```diff
@@ -231,10 +231,14 @@ func Apply(registry Registry, resources Resources) error {
 
 		existingResource, err := handler.GetRemote(resource)
 		if errors.Is(err, ErrNotFound) {
-			uid := resource.GetSpecString("uid")
-			if uid != resource.Name() {
-				return fmt.Errorf("uid '%s' and name '%s', don't match", uid, resource.GetSpecString("uid"))
+			// synthetic monitoring doesn't have uid in the spec
+			if resource.Kind() != "SyntheticMonitoringCheck" {
```
malcolmholmes, line 235: We need to fake a hidden somehow, so that every resource has one.

I have made it with `type.job`. We could make grizzly fail if the name does not match the type.job in the spec.

That way workflow.go can be the same across all handlers.

ying-jeanne: how about we check the uid exist or not, if exist, we check it is the same as type.job, otherwise ignore it. to have a fake uid would have to keep the uid updated when check method changes etc, and also i need to marshal/unmarshal the spec to get check object, it would be less evident to be maintained.

pkg/grafana/synthetic-monitoring.go
```diff
@@ -28,331 +28,236 @@ import (
...
-	req, err := http.NewRequest("GET", url, nil)
-	req.Header.Add("Authorization", "Bearer "+authToken)
-	req.Header.Add("Content-type", "application/json")
 
-	resp, err := client.Do(req)
+	return smClient, nil
+}
+
+// getProbeList retrieves the list of probe and grouped by id and name
+func getProbeList() (*Probes, error) {
```
dsotirakis, line 76: No need to be pointer? You can return `Probes{}` instead.
```suggestion
func getProbeList() (Probes, error) {
```

pkg/grafana/synthetic-monitoring.go
```diff
@@ -28,331 +28,236 @@ import (
...
-	defer resp.Body.Close()
 
-	if resp.StatusCode >= 400 {
-		body, _ := ioutil.ReadAll(resp.Body)
-		return "", fmt.Errorf("%d response while authenticating: %s", resp.StatusCode, string(body))
+	return smCheck, nil
+}
+
+// Probes allows accessing Probe objects by ID and by name
+func getType(check *synthetic_monitoring.Check) string {
```
dsotirakis, line 245: Pointer not needed here.
```suggestion
func getType(check synthetic_monitoring.Check) string {
```

pkg/grafana/synthetic-monitoring.go
```diff
@@ -28,331 +28,236 @@ import (
...
-		return "", grizzly.APIErr{Err: err, Body: data}
-
+	if check.Settings.Dns != nil {
+		return "dns"
 	}
-	return authResponse.AccessToken, nil
+	return ""
+}
+
+func getUID(check *synthetic_monitoring.Check) string {
```
dsotirakis, line 261: Pointer not needed here.
```suggestion
func getUID(check synthetic_monitoring.Check) string {
```

pkg/grafana/synthetic-monitoring.go
```diff
@@ -28,331 +28,236 @@ import (
...
 	}
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
 
-	type AuthRequest struct {
-		StackID           int `json:"stackId"`
-		MetricsInstanceID int `json:"metricsInstanceId"`
-		LogsInstanceID    int `json:"logsInstanceId"`
-	}
+	convertProbeNameToID(&resource)
```
dsotirakis, line 215:
```suggestion
	err = convertProbeNameToID(&resource)
	if err != nil {
		return err
	}
```

dsotirakis (commented):
A few comments

dsotirakis (approved):

PR #134, opened 2021-07-01
Initial content and menu

malcolmholmes:
These are substantial changes to the (new) Grizzly docs. I have deployed them already, so they can be reviewed at https://grafana.github.io/grizzly

These docs now cover to a reasonable depth what Grizzly is, and how it works. I'd like to get these docs merged ready for a 0.2.0 release.


.github/workflows/gh-pages.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
       - name: Deploy
         uses: peaceiris/actions-gh-pages@v3
-        if: github.ref == 'refs/heads/master'
+#        if: github.ref == 'refs/heads/master'
```
malcolmholmes, line 32: need to remove this before merging

dsotirakis (approved):
Loved it 🚀

PR #137, opened 2021-07-06
Move dashboard specific check to dashboard-handler

dsotirakis:
This PR is responsible for moving the condition to check the `UID` against `Resource Name`, to the `dashboard-handler`.


malcolmholmes (approved):
Great, thanks!

PR #140, opened 2021-07-08
Stop offering a delete link for expiring previews

rhowe:
Only display a link to delete previews if they aren't set to expire.

It seems unnecessary to offer a manual delete link if the user has requested automatic deletion after an expiry time.

This should result in less UI clutter and fewer unintentional deletions from hitting the wrong link by accident.


rhowe:
@dsotirakis not wanting to bug, but how about this one?

malcolmholmes (approved):
This looks reasonable to me.

PR #141, opened 2021-07-09
Ensure 'grr preview' processes all previewable resources

### LINKED ISSUE #139: "grr preview" sometimes doesn't generate snapshots for all dashboards
I see nondeterministic behaviour with `grr preview` where sometimes it only generates snapshots for a random subset of the dashboards defined in the input.

Rerunning the command will result in a different set, sometimes I have to run it upwards of 5 times before it generates a snapshot for the dashboard I wish to actually preview.

rhowe:
Instead of stopping when we encounter a resource which doesn't support previewing (e.g. a DashboardFolder), just skip it with a warning.

Fixes #139

### DIFF
--- a/pkg/grizzly/workflow.go
+++ b/pkg/grizzly/workflow.go
@@ -285,7 +285,7 @@ func Preview(resources Resources, opts *PreviewOpts) error {
 		previewHandler, ok := handler.(PreviewHandler)
 		if !ok {
 			notifier.NotSupported(resource, "preview")
-			return nil
+			continue
 		}
 		err = previewHandler.Preview(resource, opts)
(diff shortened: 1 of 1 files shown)


rhowe:
@dsotirakis not wanting to bug you, but what do you think of this?

Also we have been running grizzly with https://github.com/grafana/grizzly/pull/140 applied and find it more usable that way

dsotirakis (approved):
I think it does make sense - sorry for missing that @rhowe, LGTM!

Ready to merge if @malcolmholmes is also happy with it :)

malcolmholmes (approved):
This makes sense, thanks.

PR #151, opened 2021-08-07
Make Registry global and use notifier package

malcolmholmes:
Currently, we create a `Registry` at startup, then hand it around everywhere. As
a configuration that does not change during the life of the app, this is an
unnecessary overhead that complicates matters.

The immediate trigger for this change is the need for a function on the
`Resource` type to be able to identify its handler. This would be a pain to
have a `Resource` object have access to the `Registry` - it would make the
code unnecessarily complex. This change, on the other hand, makes the code
significantly simpler, reducing the unnecessary handing around of static
configuration

Likewise, `registry.Notifier` is a reference to a type that only has
functions. These functions are better off as functions within a package.
This again is much simpler.

### DIFF
--- a/pkg/grizzly/registry.go
+++ b/pkg/grizzly/registry.go
@@ -1,0 +1,86 @@
+package grizzly
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/gobwas/glob"
+)
+
+// Provider describes a single Endpoint Provider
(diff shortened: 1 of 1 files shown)


ying-jeanne (approved):
LGTM, i like this registry/notifier simplification. it is much easier to read.

PR #154, opened 2021-08-16
Use absolute import paths

malcolmholmes:
Due to legacy support for hidden fields, Grizzly does not load Jsonnet files directly - it loads its own Jsonnet, which imports the referenced jsonnet.

This means that the path to the `lib` and `vendor` directories will actually be the current directory, not the directory in which the loaded jsonnet file resides.

This PR makes those paths absolute and relative to the jsonnet file's directory, and fixes this issue.

### DIFF
--- a/pkg/grizzly/jsonnet.go
+++ b/pkg/grizzly/jsonnet.go
@@ -1,6 +1,8 @@
 package grizzly
 
 import (
+	"path/filepath"
+
 	"github.com/google/go-jsonnet"
 )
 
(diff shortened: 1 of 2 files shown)


dsotirakis (approved):

ywwg:
The way this is written, if the path specified in opts.JsonnetPaths is already absolute (begins with "/"), this will still join with the cwd, resulting in an invalid path. I think this can be fixed by using testing for !filepath.IsAbs before doing the operation.

malcolmholmes:
@ywwg  Fix a bug, add a bug...

Thanks for the (valid) observation. Fancy rolling a quick PR?

ywwg:
np: #156

PR #155, opened 2021-08-16
Don't use local when adding to global registry

malcolmholmes:
Silly mistake when adding global registry. Without this we get "handler not found" everywhere.


dsotirakis (approved):
Java-esque PR, LGTM!

PR #156, opened 2021-08-24
Do not modify paths that are already absolute.

ywwg:
Fix for a bug introduced in #154, which caused already-absolute paths to have the cwd prepended to them.


malcolmholmes (approved):
Thanks!

PR #159, opened 2021-08-24
Add GetUID function to resource object

malcolmholmes:
The new tests have shown up some naming issues.

We have `resource.Name()`, which extracts `metadata.name`, but this is *not*
the resource's UID, which in some cases needs to be a compound element, e.g.
`$NAMESPACE.$NAME` for a `PrometheusRuleGroup` or `$TYPE.$JOB` for a
`SyntheticMonitoringCheck`.

This PR adds a `GetUID()` function that correctly returns a UID across all
resource types.


dsotirakis (approved):
LGTM, but it would be also nice to add some unit tests, at least for the funcs that have conditions in them :)

PR #160, opened 2021-08-24
Set name metadata correctly for prom rules

malcolmholmes:
Here, `uid` is `$NAMESPACE.$NAME`. We should not set the `name` to that.

Currently, if we re-render a UID we will get `$NAMESPACE.$NAMESPACE.$NAME`.

This PR corrects this error by setting the name to just `group.name`. The
namespace is set in metadata in the next line.

### DIFF
--- a/pkg/grafana/rules.go
+++ b/pkg/grafana/rules.go
@@ -52,7 +52,7 @@ func getRemoteRuleGroup(uid string) (*grizzly.Resource, error) {
 						"rules": group.Rules,
 					}
 					handler := RuleHandler{}
-					resource := grizzly.NewResource(handler.APIVersion(), handler.Kind(), uid, spec)
+					resource := grizzly.NewResource(handler.APIVersion(), handler.Kind(), group.Name, spec)
 					resource.SetMetadata("namespace", namespace)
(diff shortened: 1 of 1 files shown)


pkg/grafana/rules.go
```diff
@@ -52,7 +52,7 @@ func getRemoteRuleGroup(uid string) (*grizzly.Resource, error) {
 						"rules": group.Rules,
 					}
 					handler := RuleHandler{}
-					resource := grizzly.NewResource(handler.APIVersion(), handler.Kind(), uid, spec)
+					resource := grizzly.NewResource(handler.APIVersion(), handler.Kind(), group.Name, spec)
```
dsotirakis, line 55: Reeeeeeally nit, but can be `name` instead of `group.Name`, since we've already extracted the var

dsotirakis (approved):
LGTM, nit comment

pkg/grafana/rules.go
dsotirakis, line 55: Reeeeeeeeally nit, but we can use `name` instead of `group.Name` since we have already extracted the var

PR #162, opened 2021-08-24
Always apply handlers in order

malcolmholmes:
Grizzly consumes code from Tanka that extracts "manifests" from a
JSON tree (i.e. output from Jsonnet script). This is returned as a
`map[string]manifest.Manifest`. As this is a map, it is unordered,
meaning resources are returned in a different order each time.

This can cause problems when ordering matters, notably when a
dashboard is applied before its folder.

This PR creates a consistent ordering for resources, they will
always be processed in consistent handler order.

Note, ordering within a handler will not (yet) be consistent, but
this shouldn't cause issues.

### DIFF
--- a/pkg/grafana/provider.go
+++ b/pkg/grafana/provider.go
@@ -27,9 +27,9 @@ func (p *Provider) APIVersion() string {
 // GetHandlers identifies the handlers for the Grafana provider
 func (p *Provider) GetHandlers() []grizzly.Handler {
 	return []grizzly.Handler{
+		NewDatasourceHandler(*p),
 		NewFolderHandler(*p),
 		NewDashboardHandler(*p),
-		NewDatasourceHandler(*p),
 		NewRuleHandler(*p),
 		NewSyntheticMonitoringHandler(*p),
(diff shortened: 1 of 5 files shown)


pkg/grafana/provider.go
```diff
@@ -27,10 +27,10 @@ func (p *Provider) APIVersion() string {
...
 func (p *Provider) GetHandlers() []grizzly.Handler {
 	return []grizzly.Handler{
+		NewDatasourceHandler(*p),
 		NewFolderHandler(*p),
 		NewDashboardHandler(*p),
-		NewDatasourceHandler(*p),
-		NewSyntheticMonitoringHandler(*p),
 		NewRuleHandler(*p),
+		NewSyntheticMonitoringHandler(*p),
 	}
```
malcolmholmes, line 35: Datasources -> Folders -> Dashboards -> Prometheus Rules -> Synthetic Monitoring seems a reasonable order of precedence.

dsotirakis (approved):
LGTM! 🚀

PR #163, opened 2021-08-27
Switch to dot as resource separator

malcolmholmes:
I mistakenly assumed that a dot was a special character in globs. It isn't - a single character glob is a question mark.

Therefore, we can revert to using dots as separators everywhere. This PR does that, but it also keeps slash separators working for command line targets.

### DIFF
--- a/pkg/grizzly/providers.go
+++ b/pkg/grizzly/providers.go
@@ -136,10 +136,12 @@ func (r *Resource) MatchesTarget(targets []string) bool {
 	if len(targets) == 0 {
 		return true
 	}
-	key := r.Key()
+	UID := r.UID()
+	slashKey := fmt.Sprintf("%s/%s", r.Kind(), UID)
+	dotKey := fmt.Sprintf("%s.%s", r.Kind(), UID)
 	for _, target := range targets {
 		g := glob.MustCompile(target)
-		if g.Match(key) {
+		if g.Match(slashKey) || g.Match(dotKey) {
 			return true
 		}
 	}
--- a/pkg/grizzly/providers.go
+++ b/pkg/grizzly/providers.go
@@ -43,6 +43,12 @@ func (r Resource) String() string {
 	return r.Key()
 }
 
+// Key returns a key that combines kind and uid
+func (r *Resource) Key() string {
+	uid := r.UID()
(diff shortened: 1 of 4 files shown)


dsotirakis (approved):
LGTM!

pkg/grizzly/providers.go
```diff
@@ -131,10 +136,12 @@ func (r *Resource) MatchesTarget(targets []string) bool {
 	if len(targets) == 0 {
 		return true
 	}
-	key := r.Key()
+	UID := r.UID()
+	slashKey := fmt.Sprintf("%s/%s", r.Kind(), UID)
+	dotKey := fmt.Sprintf("%s.%s", r.Kind(), UID)
```
samjewell, line 141: I'd be tempted to use the following here, to have less duplication and fewer sources of truth:

```suggestion
	dotKey := r.Key() // fmt.Sprintf("%s.%s", r.Kind(), UID)
	slashKey := fmt.Sprintf("%s/%s", r.Kind(), UID)
```

pkg/grizzly/providers.go
```diff
@@ -43,17 +43,22 @@ func (r Resource) String() string {
...
 	}
-	return fmt.Sprintf("%s/%s", r.Kind(), uid)
+	return uid
+}
+
+// Key returns a key that combines kind and uid
+func (r *Resource) Key() string {
+	uid := r.UID()
+	return fmt.Sprintf("%s.%s", r.Kind(), uid)
 }
```
samjewell, line 62: Could be nice to define the `.Key` method **before** the `.UID()` method, so that the file reads from higher-level methods to lower level ones. Similar to the description on [this post](https://www.reddit.com/r/ruby/comments/3ztgdo/is_there_a_proper_ruby_style_for_the_order_of/cyphi7a?utm_source=share&utm_medium=web2x&context=3).

samjewell (approved):
Very nice, thanks for this change 💯

pkg/grizzly/providers.go
malcolmholmes: done

pkg/grizzly/providers.go
malcolmholmes: done

PR #172, opened 2021-09-27
Import libraries relative to execution folder

### LINKED ISSUE #171: Additional libraries are relative to file at 'resource-path', not to current folder
The additional libraries (like 'vendor' and 'lib') are relative to the target jsonnet file being run, as introduced here: https://github.com/grafana/grizzly/pull/154.

However, this breaks running grizzly from another directory. For example, in this repo, this works:

    cd examples && grr list grr.jsonnet

This does not

    grr -J examples/vendor examples/grr.jsonnet

This also works, and probably shouldn't:

    grr -J vendor examples/grr.jsonnet

because the 'vendor' folder is in 'examples'. The -J flag works fine with absolute paths.

Running `jsonnet` against the file does work in the first two cases listed, but not in the third. Can we make the `-J` flag be relative to the directory the command is ran from?

This is on master / v0.2.0-beta1

RutgerKe:
This is to make it consistent with jsonnet behavior. Fixes #171

### DIFF
--- a/pkg/grizzly/parsing.go
+++ b/pkg/grizzly/parsing.go
@@ -100,12 +100,11 @@ func ParseJsonnet(jsonnetFile string, opts Opts) (Resources, error) {
 
 	script := fmt.Sprintf(script, jsonnetFile)
 	vm := jsonnet.MakeVM()
-	absoluteFile, err := filepath.Abs(jsonnetFile)
+	currentWorkingDirectory, err := os.Getwd()
 	if err != nil {
 		return nil, err
 	}
-	path := filepath.Dir(absoluteFile)
-	vm.Importer(newExtendedImporter(path, opts.JsonnetPaths))
+	vm.Importer(newExtendedImporter(currentWorkingDirectory, opts.JsonnetPaths))
 	for _, nf := range native.Funcs() {
 		vm.NativeFunction(nf)
 	}
--- a/examples/README.md
+++ b/examples/README.md
@@ -7,5 +7,6 @@ This directory contains examples that use this format.
 To look at the examples in this directory, execute:
 
 ```
-grr show examples/grr.jsonnet
+cd examples
+grr show grr.jsonnet
 ```


pkg/grizzly/parsing.go
```diff
@@ -98,14 +98,15 @@ var script string
 // ParseJsonnet evaluates a jsonnet file and parses it into an object tree
 func ParseJsonnet(jsonnetFile string, opts Opts) (Resources, error) {
 
-	script := fmt.Sprintf(script, jsonnetFile)
-	vm := jsonnet.MakeVM()
-	absoluteFile, err := filepath.Abs(jsonnetFile)
+	currentWorkingDirectory, err := os.Getwd()
 	if err != nil {
-		return nil, err
+		log.Fatal(err)
```
malcolmholmes, line 103: Grizzly doesn't use `log.Fatal`, can you pass back a sensible error message?

RutgerKe: Thanks, good catch. I think the command can only fail in 'non-sensible' situations :wink:, like the current working directory being deleted.

malcolmholmes:
I've been thinking about this. This actually *isn't* the intended behaviour. It may need a bit more thinking about.

In the Tanka scenario, you have a clear tree, with `environments`, `libs` and `vendor` at the root. What you actually want to do is find your way from wherever you are (current working dir) to that parent dir location and start the process from there.

That is the sort of setup we really want with Grizzly too.

However, Grizzly has operated for a long time in what I could call "hack mode", where it is being used to extract pieces of functionality from within a Tanka Jsonnet tree. In this case, arbitrary locations make sense.

I would rather, though, that Grizzly think of itself as a first class citizen alongside Tanka, and that `grr.jsonnet` files be placed within the file tree to represent entrypoints that can extract relevant resources from other Jsonnet code. This way the tree structure is respected in the same way that Tanka does.

RutgerKe:
I see, the reason I made this change is because my structure is as follows:

    lib/some-lib/dashboard.libsonnet
    lib/some-lib/mixin.libsonnet
    lib/other-lib/dependency.libsonnet
    lib/..
    vendor/..
    environment/..
    somefolder/grr.jsonnet

These dashboards are being deployed using tanka to various environments (using `environment`), but a few need to deployed to some other grafana instance, by using grizzly.

Before this PR running `grr somefolder/grr.jsonnet` fails because it will look for `somefolder/lib`, running it as `grr -J ../lib grr.jsonnet` (which was required before this PR, as it would always start looking from the grr.jsonnet file) from that folder messes up the dependencies in the other files, which are also used by tanka so cannot be changed easily. (Before the pr `grr -J ../lib grr.jsonnet` from the folder or `grr -J ../lib somefolder/grr.jsonnet` are the same)

So the goal of this was to just be able to run `tk` and `grr` from the same directory, the one with `lib`, `vendor` etc. Is that what you also want for Grizzly? Or do you mean something else? I think this did work in the first release of grizzly.

Issue #181, opened 2021-10-19
Field "Name" missing when applying datasource


tenstad:
## Error
`Non-200 response from Grafana while applying datasource prometheus: 422 Unprocessable Entity [{"fieldNames":["Name"],"classification":"RequiredError","message":"Required"}]`

## How to reproduce
In the [`examples/`](https://github.com/grafana/grizzly/tree/master/examples) folder, run `grr apply grr.jsonnet`.

## Suggested solution
Add `spec["name"] = m.Metadata().Name()` below [`spec["uid"] = m.Metadata().Name()`](https://github.com/grafana/grizzly/blob/master/pkg/grafana/datasource-handler.go#L88).

malcolmholmes:
This I believe works now.

PR #182, opened 2021-10-19
Set datasource spec.name to metadata.name

tenstad:
Closes https://github.com/grafana/grizzly/issues/181

Resolves the following error when doing `grr apply grr.jsonnet` in the example folder:  
`Non-200 response from Grafana while applying datasource prometheus: 422 Unprocessable Entity [{"fieldNames":["Name"],"classification":"RequiredError","message":"Required"}]`

`spec.name` seems to be a required datasource field. I think it makes sense to copy the name from metadata, as done with `spec.uid`.

### DIFF
--- a/pkg/grafana/datasource-handler.go
+++ b/pkg/grafana/datasource-handler.go
@@ -86,6 +86,7 @@ func (h *DatasourceHandler) Parse(m manifest.Manifest) (grizzly.Resources, error
 		}
 	}
 	spec["uid"] = m.Metadata().Name()
+	spec["name"] = m.Metadata().Name()
 	resource["spec"] = spec
 	return grizzly.Resources{resource}, nil
 }


malcolmholmes:
Actually, the right way to do this, that we should switch all resource types to follow, is to throw a (clear) error when the `metadata/name` and `spec/uid` (etc) do not match.

How does this sound? Fancy doing it?

tenstad:
Not sure if I follow. `spec/uid` is currently derived from `metadata/name`, and they thus match.

Should both `spec/name` and `spec/uid` be defind by the user, rather than derived from `metadata/name`? Additionaly, `metadata/name` and `spec/uid` must match?

tenstad:
I can agree that this might not be the best solution, since the `metadata/name` in k8s (format inspiration) is more restrictive than then Grafana's datasource name. I therefore think the example should be updated instead, adding the `spec/name` field.

seb-lean:
what is the resolution for users? How do we avoid this error?

PR #185, opened 2021-10-21
Improve Jpath handling

malcolmholmes:
Ideally, we would have a way to track down the root of a project, and from there
identify where the default jsonnet paths of `lib` and `vendor` should be found.

In place of this, we need two forms:

 * Look in the directory of the jsonnet file. This has the benefit of allowing
   local overrides of other imports
 * Look in the current directory where Grizzly was invoked. This allows for
   running Grizzly in the midst of other larger jsonnet codebases.

Without this fix, the Grizzly/Jsonnet examples don't work.

### DIFF
--- a/pkg/grizzly/jsonnet.go
+++ b/pkg/grizzly/jsonnet.go
@@ -28,9 +28,16 @@ func newFileLoader(fi *jsonnet.FileImporter) importLoader {
 	}
 }
 
-func newExtendedImporter(path string, jpath []string) *ExtendedImporter {
-	absolutePaths := make([]string, len(jpath)+1)
+func newExtendedImporter(jsonnetFile, path string, jpath []string) *ExtendedImporter {
+	absolutePaths := make([]string, len(jpath)*2+1)
 	absolutePaths = append(absolutePaths, path)
+	jsonnetDir := filepath.Dir(jsonnetFile)
+	for _, p := range jpath {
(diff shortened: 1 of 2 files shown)


pkg/grizzly/jsonnet.go
```diff
@@ -28,9 +28,16 @@ func newFileLoader(fi *jsonnet.FileImporter) importLoader {
 	}
 }
 
-func newExtendedImporter(path string, jpath []string) *ExtendedImporter {
-	absolutePaths := make([]string, len(jpath)+1)
+func newExtendedImporter(jsonnetFile, path string, jpath []string) *ExtendedImporter {
+	absolutePaths := make([]string, len(jpath)*2+1)
 	absolutePaths = append(absolutePaths, path)
+	jsonnetDir := filepath.Dir(jsonnetFile)
+	for _, p := range jpath {
```
dsotirakis, line 35: This for loop can be merged with the next one, by setting a `temp := p` var and then append both?

dsotirakis (approved):
LGTM, one comment really.

pkg/grizzly/jsonnet.go
malcolmholmes: I did think of that at first. But I want to have the two dirs relative to the jsonnet file first, then the two relative to the current directory.

Duologic (approved):

PR #192, opened 2021-11-30
docker: add jsonnet, jsonnetfmt and jsonnet-bundler

MrFreezeex:
Here is a PR that adds jsonnet, jsonnetfmt and jsonnet-bundler in the docker image. Those tools are not strictly necessary for grizzly to run of course, but since grizzly is a deployment/debugging tool I think it could be a convenient thing to have those dependencies in the docker image directly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,16 @@
 FROM golang:1.16.2
+ARG JSONNET_VERSION=v0.17.0 \
+    JB_VERSION=v0.4.0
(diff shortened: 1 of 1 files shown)


malcolmholmes (approved):
thx!

MrFreezeex:
I forgot about adding CGO_ENABLED since it's built on debian and the final image is on alpine. I tested this time and it works now, sorry about that.

PR #193, opened 2021-12-01
documenting the use of a proxy with grizzly

gonzalez:
added the use of HTTPS_PROXY to our docs


malcolmholmes:
does this actually work? I'm not aware of any HTTP proxy config within Grizzly.

gonzalez:
Go checks for the environment variable and if it is set uses the proxy. I've tested this locally and my client has used it to migrate dashboards.

malcolmholmes (approved):
Thanks for the clarification!

PR #199, opened 2022-01-21
Make name to uid conversion more consistent with grizzly.jsonnet 

MrFreezeex:
So far grizzly enforce that the name strictly equals to the uid and even
set the UID in the parse function.

This is problematic because some characters are allowed in the name
and not in the uid. For instance many mixins put a ".json" has a suffix
in the name and the "." character is not something allowed in an UID.

With this commit grizzly will now only enforce that the uid exist and is
not an empty string. It also set by default to the md5 of the name in
the parse function.

Signed-off-by: Arthur Outhenin-Chalandre <arthur.outhenin-chalandre@cern.ch>

### DIFF
--- a/pkg/grafana/dashboard-handler.go
+++ b/pkg/grafana/dashboard-handler.go
@@ -3,6 +3,7 @@ package grafana
 import (
 	"fmt"
 	"path/filepath"
+	"regexp"
 
 	"github.com/grafana/grizzly/pkg/grizzly"
 	"github.com/grafana/grizzly/pkg/grizzly/notifier"
@@ -21,13 +22,17 @@ func NewDashboardHandler(provider Provider) *DashboardHandler {
 	}
 }
 
+func formatUID(name string) string {
+	var pattern = regexp.MustCompile(`[^a-zA-Z0-9_-]`)
+	return pattern.ReplaceAllString(name, "")
+}
+
 // Validate returns the uid of resource
 func (h *DashboardHandler) Validate(resource grizzly.Resource) error {
-	uid, exist := resource.GetSpecString("uid")
-	if exist {
-		if uid != resource.Name() {
-			return fmt.Errorf("uid '%s' and name '%s', don't match", uid, resource.Name())
-		}
+	uid, _ := resource.GetSpecString("uid")
+
+	if uid != formatUID(resource.Name()) {
+		return fmt.Errorf("uid '%s' and name '%s', don't match", uid, resource.Name())
 	}
 	return nil
 }
@@ -70,7 +75,7 @@ func (h *DashboardHandler) ResourceFilePath(resource grizzly.Resource, filetype 
 // Parse parses a manifest object into a struct for this resource type
 func (h *DashboardHandler) Parse(m manifest.Manifest) (grizzly.Resources, error) {
 	resource := grizzly.Resource(m)
-	resource.SetSpecString("uid", resource.GetMetadata("name"))
+	resource.SetSpecString("uid", formatUID(resource.Name()))
 	if !resource.HasMetadata("folder") {
 		resource.SetMetadata("folder", dashboardFolderDefault)
 	}
@@ -103,9 +108,8 @@ func (h *DashboardHandler) GetByUID(UID string) (*grizzly.Resource, error) {
 
 // GetRemote retrieves a dashboard as a resource
 func (h *DashboardHandler) GetRemote(resource grizzly.Resource) (*grizzly.Resource, error) {
-	uid, _ := resource.GetSpecString("uid")
-	if uid != resource.Name() {
-		return nil, fmt.Errorf("uid '%s' and name '%s', don't match", uid, resource.Name())
+	if err := h.Validate(resource); err != nil {
+		return nil, err
 	}
 	return getRemoteDashboard(resource.Name())
 }


malcolmholmes:
The UID and the name are, in the case of dashboards, intended to be essentially the same thing. If you are using mixins, and this is causing you problems, it will be because the mixin is not providing a UID within the dashboard. This in itself is problematic as Grafana won't know how to uniquely identify the file.

Here's where this case is currently handled: https://github.com/grafana/grizzly/blob/978538c210642a60399f54f208e9528764a1ce68/pkg/grizzly/grizzly.jsonnet#L28-L31

Looking at how prometheus-ksonnet handles this problem, it would make sense to just take the md5sum of the filename if no UID is provided, in the above `grizzly.jsonnet` snippet. This would solve your problem.

https://github.com/grafana/jsonnet-libs/blob/5a128df878434da37969b811e99bb9cd0a3779e3/prometheus-ksonnet/grafana/dashboards.libsonnet#L20

malcolmholmes:
There's a few things to clear up here. Firstly, the way that mixins specify their content is kinda broken. They **expose** their content to the outside world via **hidden** elements. There's no way this can be a good idea. And it causes all sorts of complications.

Grizzly has moved away from specifying resources in this way, to a Kubernetes inspired explicit method. In this world, the name and the UID must match. That's something the user can control.

Now, what you are talking about is about backwards compatibility with the mixin/hidden elements approach. This happens in the `pkg/grizzly/grizzly.jsonnet` file, which aims to transform hidden element based Jsonnet into k8s style Jsonnet. That is the *only* place where your JSON keys have any form of existence (they have no meaning in the k8s world).

So really, if this patch is going to happen, it *needs* to be focusing on `grizzly.jsonnet` as a translation, rather than on core Grizzly functionality that has deliberately moved on from the hidden element approach.

MrFreezeex:
I can do a patch for jsonnet but this also is a problem for the yaml/kubernetes style as well:
```
$ cat dashboard-simple.yaml
kind: Dashboard
metadata:
    name: not-valid.json
    folder: General
spec:
    schemaVersion: 17
    tags:
        - templated
    timezone: browser
    title: Should not be here
    uid: will-be-replaced
$ ./grr apply dashboard-simple.yaml
2022-01-26 10:21:04.390145 I | 10:21:04 Non-200 response from Grafana while applying dashboard not-valid.json: 400 Bad Request {"message":"uid contains illegal characters"}
```

malcolmholmes:
Ahh yes, I remember now. The correct behaviour should, in k8s land, be:
* if no UID, use name
* if UID and name differ, refuse to proceed, with a clear error message.

Clearly I never got around to implementing that (was assuming I had). Does it make sense? Fancy taking a stab?

MrFreezeex:
> Ahh yes, I remember now. The correct behaviour should, in k8s land, be:
> 
>     * if no UID, use name
> 
>     * if UID and name differ, refuse to proceed, with a clear error message.
> 
> 
> Clearly I never got around to implementing that (was assuming I had). Does it make sense? Fancy taking a stab?

Well this is almost what is implemented at least it was clearly intended to be this way apart that the `Parse` function always override the uid by the name. But even if I fix that it won't solve the underlying issue which is that not every character in the name could be in the uid...

If I don't have the line `uid: will-be-replaced` the error would be the same because grizzly will make the uid equals `not-valid.json` and this will fail. So there should be a function to translate the name to a uid and IMO a md5 should do the job perfectly...

MrFreezeex:
> If I don't have the line `uid: will-be-replaced` the error would be the same because grizzly will make the uid equals `not-valid.json` and this will fail. So there should be a function to translate the name to a uid and IMO a md5 should do the job perfectly...

I did that in my latest commit so it's the same behavior that you described except that I use `uid=md5(name)` instead of `uid=name`.

malcolmholmes:
but the problem you are trying to solve, a dot in a filename, is inherently derived from the mixin world (i.e. the hidden variable world). Therefore, the fix should be in the hidden->k8s translation layer, which is `grizzly.jsonnet`. Once we're in k8s resource world, if there's a dot in your name, well, remove it! In the mixin case, the filename should already have been converted to an md5sum.

MrFreezeex:
> but the problem you are trying to solve, a dot in a filename, is inherently derived from the mixin world (i.e. the hidden variable world). Therefore, the fix should be in the hidden->k8s translation layer, which is `grizzly.jsonnet`. Once we're in k8s resource world, if there's a dot in your name, well, remove it! In the mixin case, the filename should already have been converted to an md5sum.

Oh actually looking at grizzly.jsonnet it should already handle that but the k8s layer disregard that by setting uid in all cases. I will check how to make the things that grizzly.jsonnet enforced be respected 👍.

MrFreezeex:
Hmmm while checking more about this grizzly.jsonnet story I saw that I am in fact not using that but the lib in jsonnet-lib instead https://github.com/grafana/jsonnet-libs/tree/master/grizzly.

The issue is that the jsonnet-libs doesn't do any fancy uid/name conversion https://github.com/grafana/jsonnet-libs/blob/master/grizzly/resource.libsonnet whereas the jsonnet code embed in grizzly does https://github.com/grafana/grizzly/blob/master/pkg/grizzly/grizzly.jsonnet#L8 https://github.com/grafana/grizzly/blob/master/pkg/grizzly/grizzly.jsonnet#L16.

So is using the jsonnet-libs version is the right thing to do in your opinion?
Should we do the we do the same name->uid conversion as https://github.com/grafana/grizzly/blob/master/pkg/grizzly/grizzly.jsonnet#L16 in the grizzly golang code for the kubernetes style?

malcolmholmes:
Using the `jsonnet-libs` version is better, as it puts you in control much more. I'm totally happy to accept PRs against that lib too. Please do port useful functionality from `grizzly.jsonnet` over to there.

MrFreezeex:
> Using the `jsonnet-libs` version is better, as it puts you in control much more. I'm totally happy to accept PRs against that lib too. Please do port useful functionality from `grizzly.jsonnet` over to there.

Thanks I have created https://github.com/grafana/jsonnet-libs/pull/735.

I also updated some code here to better reflect what grizzly.jsonnet is doing to convert name to uid (https://github.com/grafana/grizzly/blob/master/pkg/grizzly/grizzly.jsonnet#L16). It will mostly not be a concern for me with the new PR to jsonnet-libs though so if you don't want something like this feel free to close it.

MrFreezeex:
> I also updated some code here to better reflect what grizzly.jsonnet is doing to convert name to uid (https://github.com/grafana/grizzly/blob/master/pkg/grizzly/grizzly.jsonnet#L16). It will mostly not be a concern for me with the new PR to jsonnet-libs though so if you don't want something like this feel free to simply close the PR.

I suppose the same problem also hold for dashboards folders (like space in the name won't probably work right now).

Issue #209, opened 2022-04-16
Anonymous authentication no longer works


v-zhuravlev:
Hi, One thing I noticed, that if I'm not mistaken this [PR](https://github.com/grafana/grizzly/pull/197) makes authentication mandatory (which was optional in fact before that):
https://github.com/grafana/grizzly/pull/197/files#diff-6f38156d91a983d8aa26b50d3fb6dfb5764ea755a9cd5b7da70258e9be77cf92L31

so `grr` no longer works because token is always required, even if
 [anonymous access](https://grafana.com/docs/grafana/latest/auth/grafana/#anonymous-authentication) (which is quite useful for local dev environments, for example) is enabled.

```
# only GRAFANA_URL set in env:
wabbit@vzhuravl-mac node-mixin % env | grep GRAF
GRAFANA_URL=http://grafana.k3d.localhost:9999

# latest build:
wabbit@vzhuravl-mac node-mixin % ../../../grizzly/cmd/grr/grr  watch .  mixin.libsonnet
INFO[0000] Watching for changes                         
INFO[0001] Changes detected. Applyingmixin.libsonnet    
INFO[0001] Applying 6 resources                         
ERRO[0001] Error: Require GRAFANA_TOKEN                 

# version v0.2.0-beta2:
wabbit@vzhuravl-mac node-mixin % grr  watch .  mixin.libsonnet                         
2022-04-16 22:42:26.995077 I | 22:42:26 Watching for changes
2022-04-16 22:42:27.560048 I | 22:42:27 Changes detected. Applying mixin.libsonnet
Dashboard.nodes updated
Dashboard.node-multicluster-rsrc-use updated
Dashboard.node-rsrc-use updated
Dashboard.nodes-darwin updated
```
@PatMis16

v-zhuravlev:
Fixed in #210

PR #210, opened 2022-04-16
Fix access to Grafana without any auth

v-zhuravlev:
Here is the suggested fix for #209 
to allow anonymous access to grafana.


malcolmholmes (approved):
Looks clean and straightforward. Thanks for the patch.

PR #211, opened 2022-04-16
Improve jsonnet parser for rules

v-zhuravlev:
Hi! Even though it says deprecated for using hidden fields, we still using it extensively for monitoring-mixins development, especially in `grr watch` mode. 

It works great for 'dashboards only' mixins that way, but unfortunately, it stops working when rules are present, error is recevied:

```
2022-04-17 02:34:04.863636 I | 02:34:04 Error:  RUNTIME ERROR: Unexpected type string, expected number
        mixin.libsonnet:69:28-65
        mixin.libsonnet:75:17-31        thunk from <thunk from <function <fromMap>>>
        mixin.libsonnet:8:58-62 thunk from <thunk from <thunk from <object <anonymous>>>>
        mixin.libsonnet:8:43-76 thunk from <thunk from <object <anonymous>>>
        mixin.libsonnet:8:28-90 thunk from <object <anonymous>>
        mixin.libsonnet:8:13-103        object <anonymous>
        During manifestation
``` 

This is due grizzly.jsonnet [expects](https://grafana.github.io/grizzly/hidden-elements/) rules to be wrapped in the namespace:

```
  prometheusAlerts+:: {
    grizzly_rules: { <--namespace
      groups: [{
        name: 'my_rules_group',
        rules: [
          $.alert,
        ],
      }],
    },
  },
````

while monitoring-mixins by convention have alerts defined without one:

```
  _config+:: {
  ...
  },
  grafanaDashboards+:: {
  ...
  },
  prometheusRules+:: {
  ...
  },
  prometheusAlerts+:: {
      groups: [{
        name: 'my_rules_group',
        rules: [
          $.alert,
        ],
      }],
  },
```

(some examples: 
https://github.com/prometheus/node_exporter/blob/master/docs/node-mixin/alerts/alerts.libsonnet
https://github.com/prometheus/prometheus/blob/main/documentation/prometheus-mixin/alerts.libsonnet
)

So here is the suggested improvement for jsonnet rule parser that:

- Parses rule groups even if they are not wrapped into namespace. This is extremely useful when working with monitoring-mixins. 'grizzly_rules' namespace is added
- Loads all rule groups instead of the first one.

This fix not only allows loading of monitoring-mixins dashboards inside grafana **without any grizzly specific changes to original mixin**, but also can load monitoring-mixins alerts into cortex as well:

```
user@vzhuravl-mac node-mixin % ../../../grizzly/cmd/grr/grr  watch .  mixin.libsonnet 
INFO[0000] Watching for changes                         
INFO[0001] Changes detected. Applyingmixin.libsonnet    
INFO[0001] Applying 8 resources                         
Dashboard.node-rsrc-use updated
Dashboard.nodes-darwin updated
Dashboard.nodes updated
Dashboard.node-multicluster-rsrc-use updated
PrometheusRuleGroup.grizzly_rules.node-exporter.rules no differences
PrometheusRuleGroup.abc.node-exporter-small no differences
PrometheusRuleGroup.zzz.node-exporter-small2 no differences
PrometheusRuleGroup.zzz.node-exporter no differences
```

image

### DIFF
--- a/docs/content/hidden-elements.md
+++ b/docs/content/hidden-elements.md
@@ -1,5 +1,5 @@
 ---
-date: "2021-06-28T00:00:00+00:00"
+date: "2022-04-17T00:00:00+00:00"
 title: "Hidden Elements"
 ---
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/grizzly/grizzly.jsonnet [hunk 1]
```diff
@@ -65,21 +65,32 @@ local convert(main, apiVersion) = {
...
+              g.name,
+              spec={
+                rules: g.rules,
+              },
+              metadata={ namespace: ns }
+            )
+            for g in allNamespaced[ns].groups
+          ]
+          for ns in std.objectFields(allNamespaced)
+        ] else [];
```
malcolmholmes, line 93: if `key in main` is true, then doesn't this result in an array within an array?

docs/content/hidden-elements.md
```diff
@@ -82,9 +82,52 @@ Prometheus alerts and recording rules can be defined too, for example:
   },
 }
 ```
-Here, we first create an alert rule element (`grizzly_alert`), which we then add to `prometheusAlerts`. `PrometheusAlerts` is where Grizzly expects to find alerts for Prometheus configured when using 'hidden
+Here, we first create an alert rule element (`grizzly_alert`), which we then add to namespace `grizzly_rules` inside `prometheusAlerts`. `PrometheusAlerts` is where Grizzly expects to find alerts for Prometheus configured when using 'hidden
 elements.
```
malcolmholmes, line 86:
```suggestion
elements'.
```

malcolmholmes (commented):
Minor comments, looks otherwise done well, thanks!

pkg/grizzly/grizzly.jsonnet
v-zhuravlev: Good point! Still, grizzly tolerated that :) Updated to use nested loop without extra array in the result.

malcolmholmes (approved):
Looks great, thanks for this.

PR #218, opened 2022-07-05
Add support for Grafana Library Panels

malcolmholmes:
Adds support for Grafana library panels to Grizzly.

Also starts a transition to using the Golang client for Grafana. When
Grizzly was started, this was not sufficiently mature. This is no longer
the case. A transition to the golang client should reduce the amount
of code required in Grizzly.

### DIFF
--- a/pkg/grafana/library-panel-handler.go
+++ b/pkg/grafana/library-panel-handler.go
@@ -1,0 +1,168 @@
+package grafana
+
+import (
+	"fmt"
+	"path/filepath"
+
+	grafana "github.com/grafana/grafana-api-golang-client"
+	"github.com/grafana/grizzly/pkg/grizzly"
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
+)
+
+// LibraryPanelHandler is a Grizzly Handler for Grafana library panel
+type LibraryPanelHandler struct {
+	Provider Provider
+}
+
+// NewLibraryPanelHandler returns a new Grizzly Handler for Grafana library panel resources
+func NewLibraryPanelHandler(provider Provider) *LibraryPanelHandler {
+	return &LibraryPanelHandler{
+		Provider: provider,
+	}
+}
+
+// Kind returns the kind for this handler
+func (h *LibraryPanelHandler) Kind() string {
+	return "LibraryPanel"
+}
+
+// Validate returns the uid of resource
+func (h *LibraryPanelHandler) Validate(resource grizzly.Resource) error {
+	uid, exist := resource.GetSpecString("uid")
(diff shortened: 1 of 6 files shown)


pkg/grafana/library-panel-handler.go
```diff
@@ -0,0 +1,168 @@
...
+	grafana "github.com/grafana/grafana-api-golang-client"
+	"github.com/grafana/grizzly/pkg/grizzly"
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
+)
+
+// LibraryPanelHandler is a Grizzly Handler for Grafana library panel
+type LibraryPanelHandler struct {
+	Provider Provider
+}
+
```
hsyed-dojo, line 16: nit: a type assertion here would be nice to assert that `LibraryPanelHandler` conforms to `grizzly.Handler`.

hsyed-dojo (approved):
LGTM

ccjones88:
So I pulled this down locally to try it out and I'm kind of caught in this catch 22 of both dashboards using the library panel and the library panel itself saying that it cannot be found.

Steps for reproducing:

1. Have two dashboards sharing 1 or more library panels
2. grr pull library panels/dashboards ( I myself am using `grr pull -d resources -t "LibraryPanel/*" -t "Dashboard/*" -l debug` )
3. Next grr apply to a Grafana instance that does not have those dashboards/library panels already existing. I've tried quite a few different ways here:

- `grr apply -d resources -t "Dashboard/*" -t  "LibraryPanel/*" -l debug`
- `grr apply -d resources -t  "LibraryPanel/*" -t "Dashboard/*" -l debug`
- `grr apply -d resources -t  "LibraryPanel/SpecificUID" -l debug`
- `grr apply -d resources -t "Dashboard/SpecificUID" -l debug`
- `grr apply -d resources -t  "LibraryPanel/SpecificUID" -t "Dashboard/SpecificUID" -l debug`

Dashboard apply always ends up with `{"message":"Error while connecting library panels"}`
LibraryPanel apply has `{"message":"library element could not be found"}`
Server Side logs show `level=error msg="Error while connecting library panels" error="library element could not be found"` corresponding to grr apply for the Dashboard and `level=error msg="library element could not be found" error="library element could not be found"` corresponding to grr apply for Library Panels

I'm hoping I'm just missing something obvious, but figured it'd be beneficial to let you all know now before it's merged in case there is a further bug here. Let me know if I can add any more info. 

Grafana `v9.1.5`
grr version `dev-v0.2.0-1-g5f400a8-dirty`

theSuess:
Resolved in #282

PR #221, opened 2022-08-02
Adding logic to handle grafana alerting ruler type

mikkergimenez:


malcolmholmes:
Hi. Curious how come you closed this PR. Looks interesting.

mikkergimenez:
Hello @malcolmholmes Closed because it's not polished ready for primetime, but happy to re-open it to work out in the open on it.  I am ready for some kinda feedback as to whether or not this is heading in the right direction, some of it feels a bit hacky.

malcolmholmes:
I really want to see this in Grizzly! Looking at your impl, it looked reasonable. However, if I were to do it now, I would use the grafana api golang client which now has support for alerts. A PR for library panels in Grizzly uses the client. Could you make that work?

PR #223, opened 2022-08-29
1 disable tls verify in https connections

luislain-adi:


malcolmholmes:
If you can make this configurable, we can consider it. How does https://tanka.dev handle it?

malcolmholmes:
Closing as stale: the HTTP client has been replaced by the new Grafana OpenAPI client so this would need to be done differently (assuming the new client doesn't already support it).

PR #247, opened 2023-06-12
Fix marshalling scalar blocks

### LINKED ISSUE #245: `pull` command is using `|4-` notation in yaml, which is not compatible with the `list` or `apply` commands
If the query starts with an blank line, `grr pull` will generate a yaml file using the `|4-` notation. Although this is valid yaml, the `list` and `apply` commands do not work when this notation is present in the yaml files.

Query that starts with a blank line:
image

Will result in this expression when using `grr pull`
```
              expr: |4-
                rate(
                    node_cpu_seconds_total{mode="system"}[$__rate_interval]
                )
```

In this case `grr list` will fail:
```
grr list resources/dashboards/general/dashboard-mwwg8CQVk.yaml
INFO[0000] Listing 0 resources
API VERSION    KIND    UID
```

If I simply change `|4-` to `|-` in the yaml and rerun the `list` command:
```
grr list resources/dashboards/general/dashboard-mwwg8CQVk.yaml
INFO[0000] Listing 1 resources
API VERSION                     KIND         UID
grizzly.grafana.com/v1alpha1    Dashboard    mwwg8CQVk
```

Version info:
Grizzly: 2023-06-02 09:33:31.610732 I | grr version v0.2.0
Grafana: 9.2.15

K-Phoen:
Fixes #245

There is currently a bug in [go-yaml/yaml](https://github.com/go-yaml/yaml) causing indentation hints on scalar blocks to be set incorrectly. As a result, they can not be unmarshalled.

[A PR has been open upstream](https://github.com/go-yaml/yaml/issues/643) a few years ago but hasn't seen any meaningful activity recently.

This PR is an attempt at fixing this (un)marshalling issue by using an alternative, regurlarly maintained, YAML library: [goccy/go-yaml](https://github.com/goccy/go-yaml)

Some quick manual testing showed that goccy/go-yaml resolves #245, but there is more validation needed to decide if switching to a new dependency is viable:

* goccy/go-yaml is highly likely to marshal resources differently. While some degree of change is *probably* acceptable, I currently don't know how different marshalled resources are.
* there isn't any unit test covering the issue, I will add one.


K-Phoen:
The marshaled resources are actually fairly similar with the new library. The main differences come from inconsistencies in [go-yaml/yaml](https://github.com/go-yaml/yaml).

For example, it would output:

```yaml
    panels:
        - id: 4
          gridPos:
            h: 22
            w: 10
            x: 14
            "y": 0
          tags:
            - hosted-metrics
```

(note the inconsistent indentation: mostly 4 spaces but sometimes 2 spaces)

While [goccy/go-yaml](https://github.com/goccy/go-yaml) generates:

```yaml
    panels:
        - id: 4
          gridPos:
              h: 22
              w: 10
              x: 14
              "y": 0
          tags:
              - hosted-metrics
```

(consistent, 4 spaces indentation)

joanlopez:
Hi @K-Phoen,

I tried to `grr pull` and `grr diff` resources (i.e. dashboards) from a Grafana instance, and while I'd expect no differences, I see a few of them, that look like:

```
Dashboard.hOipiybnz changes detected:
--- Remote
+++ Local
@@ -292,7 +292,7 @@
                 language: plaintext
                 showLineNumbers: false
                 showMiniMap: false
-            content: |4-
+            content: |-
                 <h1 style="text-align: center;">Docker</h1>
                 <h3 style="text-align: center;"><a href="/d/MQ5mdQbnk/docker-monitoring">More</a></h1>
             mode: html
```

It also looks like similar differences are detected when following the same steps with the current version (using `go-yaml/yaml`). So, do you think these are acceptable? Or should these completely disappear with the lib replacement?

<br/>

Other than that, is there any other test that you would suggest doing before merging this PR? (cc/ @malcolmholmes)
I'd be happy to help with getting this merged. So far, I've merged latest changes from `master` to solve conflicts.

Thanks!

malcolmholmes:
The thing that stopped me merging this PR is the fact that it results in whitespace (and other) no-op diffs. Any users that are using diff to detect meaningful changes would find a whitespace change annoying.

And given that there's a simple solution (don't lead your queries with a newline), I didn't tackle the question of whether this was a valid/valuable change.

I'm totally open to using a different approach to YAML marshalling, I just want to know its enough to justify any discomfort users may experience.

malcolmholmes:
Interesting questions - you're totally on the mark here. A JSON diff would be amazing, actually. And whitespace changes are less of a bad thing in PRs nowadays due to clever diff highlighting whitespace differently.

malcolmholmes:
FTR: This PR improves matters but does not resolve them. Grafana dashboards can do some weird things with their JSON. Here's some issues that show up with this library when parsing and rendering dashboards found in the wild: https://github.com/goccy/go-yaml/pull/418

malcolmholmes:
converted to draft as there's nothing final here

malcolmholmes:
Unfortunately, this library introduces a separate set of issues caused by the new library. I'm closing it for now. Should this library resolve its issues, we can re-open/reconsider. Keeping #245 open to keep track of the original issue.

PR #251, opened 2023-07-24
Update dependencies to latest minor

joanlopez:
While trying to contribute to Grizzly (_another pull request is coming_), I found different issues with some of the dependencies (unreachable versions/paths, etc), so I decided it could be a good opportunity to update the project dependencies to their latest minor. Few of them have a new major available, but I preferred to keep the risk of breaking anything low.

Indeed, that's my main question now: Is there any test process to verify such changes? 
So far, I tried with `make test` and it worked, but I'm wondering if that's reliable enough or if there's anything else to verify.

Thanks!


joanlopez:
Argh! I suspect that as [I've updated Go-Jsonnet dependency](https://github.com/grafana/grizzly/blob/joanlopez/update-deps/go.mod#L10), I'd probably need to update [the corresponding Dockerfile `ARG`](https://github.com/grafana/grizzly/blob/master/Dockerfile#L2), right? 🤔

PR #255, opened 2023-09-29
Add support for pulling and pushing dashboard json

undef1nd:
This PR introduces `-s` flag for `pull` and `apply` commands that allows pulling/pushing dashboard JSON as is rather than the Grizzly resource definition. 

The `apply -s` command with currently requires a folder UID as a parameter, limiting the ability to push multiple dashboards to different folders in one go. By default, resources are pushed to the `general` folder if no folder UID is specified.


cmd/grr/workflow.go
```diff
@@ -71,6 +72,23 @@ func pullCmd() *cli.Command {
 	}
 	return initialiseCmd(cmd, &opts)
 }
+
+func pullDashboardCmd() *cli.Command {
+	cmd := &cli.Command{
+		Use:   "pull-dashboard <resource-path>",
```
joanlopez, line 78: Two questions:

- Do we want a completely new command for this?
- Would make sense to extend this new feature to other resources?

I feel like this could be part of one of the existing commands (without adding breaking changes), so I'd like to evaluate that alternative as well.

cmd/grr/workflow.go
```diff
@@ -71,6 +72,23 @@ func pullCmd() *cli.Command {
...
+		Short: "Pulls remote dashboards and writes them to local source in JSON format",
+		Args:  cli.ArgsExact(1),
+	}
+	opts := grizzly.Opts{
+		Legacy: true,
+	}
+
+	cmd.Run = func(cmd *cli.Command, args []string) error {
+		return grizzly.Pull(args[0], opts)
```
joanlopez, line 87: Wouldn't this pull more resources and not only dashboards? I feel like this piece of code should manipulate `grizzly.Opts.Targets` before calling `grizzly.Pull` to get the expected behavior. 

Note that the user may also provide it, to refer specific dashboards.

joanlopez: Perhaps using `targetsOfKind` also here? 🤔

pkg/grizzly/config.go
```diff
@@ -11,6 +11,10 @@ type Opts struct {
 	Directory    bool
 	JsonnetPaths []string
 	Targets      []string
+
+	// Used for supporting commands that output dashboard JSON
+	Folder string
+	Legacy bool
```
joanlopez, line 17: I'm not super convinced about "legacy", but I don't have any super convincing alternative. What about something more specific, like SpecOnly? Or something like that, that states more clearly that it just manipulates the "spec" part?

In fact, I don't think this is legacy (yet, at least).

.gitignore
```diff
@@ -5,3 +5,6 @@
 *~
 *.orig
 *.swp
+
+# Ignore IDE files
+.idea
```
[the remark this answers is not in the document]
undef1nd: oh good comment, I missed vscode being in the list already

cmd/grr/workflow.go
undef1nd: I agree. I mentioned it in the PR description: I'm considering adding a flag to already existing commands instead of adding new commands.

>Would make sense to extend this new feature to other resources?

Maybe. Suggest to discuss it offline together with @malcolmholmes

cmd/grr/workflow.go
```diff
@@ -125,6 +143,46 @@ func applyCmd() *cli.Command {
...
+		}
+
+		resources, err := grizzly.Parse(args[0], opts)
+		if err != nil {
+			return err
+		}
+		return grizzly.Apply(resources)
+	}
+
+	cmd.Flags().StringVarP(&opts.Folder, "folder", "f", "general", "folder to push dashboards to")
```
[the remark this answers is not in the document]
undef1nd: I've added a task for this to our backlog, so we don't forget to make a decision reg this

cmd/grr/workflow.go
undef1nd: Yes-yes, good catch!

cmd/grr/workflow.go
```diff
@@ -110,21 +120,49 @@ func diffCmd() *cli.Command {
 func applyCmd() *cli.Command {
 	cmd := &cli.Command{
 		Use:   "apply <resource-path>",
-		Short: "apply local resources to remote endpoints",
+		Short: "apply local resources to remote endpoints, -s flag is only applicable to dashboards",
```
joanlopez, line 123: Nit; I think it's enough if this just sits next to the flag explanation

pkg/grizzly/workflow.go
```diff
@@ -103,7 +103,7 @@ func ListRemote(opts Opts) error {
 	return w.Flush()
 }
 
-// Pulls remote resources
+// Pull pulls remote resources, if opts.JSONSpec is true, saves them as JSON files
```
joanlopez, line 106: I'd clarify it just save the spec, otherwise it may seem that it's only the wire format what changes.

pkg/grizzly/parsing.go
```diff
@@ -166,12 +247,24 @@ func MarshalYAML(resource Resource, filename string) error {
 	if err != nil {
 		return err
 	}
+	return writeFile(filename, []byte(y))
+}
+
+func MarshalJSON(resource Resource, filename string) error {
```
joanlopez, line 253: I'd either:
- Rename it to something more specific.
- Or change the signature to just expect the `Spec`.

Otherwise, it may lead to confusions, cause within the Grizzly package there would be two very similar methods (`MarshalJSON` & `MarshalYAML`) with the same signature, that do different things (not only using either JSON or YAML as data format).

cmd/grr/workflow.go
```diff
@@ -110,21 +120,49 @@ func diffCmd() *cli.Command {
...
+
 		resources, err := grizzly.Parse(args[0], opts)
 		if err != nil {
 			return err
 		}
 		return grizzly.Apply(resources)
 	}
+
+	cmd.Flags().StringVarP(&opts.FolderUID, "folder", "f", generalFolderUID, "folder to push dashboards to")
+	cmd = initialiseSpecOnly(cmd, &opts)
```
joanlopez, line 141: Nit; Even if re-used in different commands, I'd prefer this be inlined (easier to read, less jumps), but I leave it up to you!

joanlopez (approved):

PR #257, opened 2023-10-05
Set up Grafana client for Go

joanlopez:
It set ups the baseline to start transitioning from a manually written client for Grafana to https://github.com/grafana/grafana-api-golang-client, which will make the maintenance burden of the whole project lower, and will keep it aligned with improvements added into the shared one.

It also:
- A refactor of how the client is initialized, in order to inject it into the provider and get rid of global variables:
  - This is kinda requirement for future work (like support for multiple contexts, like in k8s).
  - Once completely replaced, it will stop breaking CI due to data races on environment variables used by different tests. 
- Includes an example of how to use the client, in this case to get a single data source by uid.

It should not break the current behavior. Similar ideas introduced in https://github.com/grafana/grizzly/pull/218.

Thanks!


pkg/grafana/datasource-handler.go
```diff
@@ -133,3 +136,28 @@ func (h *DatasourceHandler) Add(resource grizzly.Resource) error {
 func (h *DatasourceHandler) Update(existing, resource grizzly.Resource) error {
 	return putDatasource(resource)
 }
+
+func (h *DatasourceHandler) getRemoteDatasource(uid string) (*grizzly.Resource, error) {
+	ds, err := h.Provider.client.DataSourceByUID(uid)
+
+	var nf gapi.ErrNotFound
+	if err != nil && errors.As(err, &nf) {
+		ds, err = h.Provider.client.DataSourceByUID(uid) // TODO: Replace by DataSourceByName
```
joanlopez, line 145: Waiting for https://github.com/grafana/grafana-api-golang-client/pull/164 to be merged

pkg/grafana/datasource_test.go
```diff
@@ -2,39 +2,46 @@ package grafana
...
+
 	t.Run("get remote datasource - success", func(t *testing.T) {
-		resource, err := getRemoteDatasource("AppDynamics")
+		resource, err := handler.GetByUID("AppDynamics")
 		require.NoError(t, err)
 
 		require.Equal(t, resource.APIVersion(), "grizzly.grafana.com/v1alpha1")
 		require.Equal(t, resource.Name(), "AppDynamics")
-		require.Len(t, resource.Spec(), 18)
+		require.Len(t, resource.Spec(), 13)
```
joanlopez, line 36: Zero values are now ignored (`database`, `jsonData`, `secureJsonFields`, etc), but the behavior should remain the same, as these are non-required by the HTTP API (so once `apply`-ed back, should be equivalent).

pkg/grafana/utils.go
```diff
@@ -27,3 +28,17 @@ func extractFolderUID(d DashboardWrapper) string {
...
+		return nil, err
+	}
+
+	var result map[string]interface{}
+	if err := json.Unmarshal(jsonData, &result); err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
```
malcolmholmes, line 44: I thought https://github.com/mitchellh/mapstructure was in use somewhere in Grizzly. Seems not, but it does the same.

malcolmholmes:
Looks good to me, when all handlers are done!

pkg/grafana/utils.go
joanlopez: I guess I'm biased by my early days in Go, when the community was generally more in favor (in comparison to other communities like JS) to avoid 3rd party dependencies just for few lines of code, so I didn't even consider it.

joanlopez: However, after taking a look, I think it wouldn't be as simple as that.

These few lines of code are "re-using" the `json` tags defined in the client. But, https://github.com/mitchellh/mapstructure has its own tags (otherwise it would store them capitalized). 

So, I'd either need to add these flags in the client (I don't like it, tho I'm open for discussions 😄) or create the corresponding DTOs in Grizzly and map them before _decoding_, which is... even more lines of code and noise? 🤔 

Wdyt?

mbarrien:
FYI, I have a fork that I made independently in the past few weeks that changes all the handlers over to the Golang API at https://github.com/mbarrien/grizzly/tree/grafana-golang-api (and even adds support for alert rules by using the branch from https://github.com/grafana/grafana-api-golang-client/pull/144). I've been hesitant to PR it because so many fields change due to all the omitempty's in the JSON dump of the official grafana API that it causes large amounts of diffs.

However I can verify that for my purposes my fork is working. I could PR it and get this all done?

mbarrien:
I've also been hesitant to make the PR since it seems like Grafana is trying to move away from that handwritten Golang library in favor of one generated from OpenAPI at https://github.com/grafana/grafana-openapi-client-go (status being tracked in https://github.com/grafana/grafana/issues/47827). It feels like if that OpenAPI library ever actually gets finished, it'll require yet another switch in libraries. The grafana-api-golang-client library already has a message in the top level README already implies that it isn't a priority to update, only so far as getting the Terraform provider working.

It would be helpful if a Grafana employee could weigh in on what the state of the OpenAPI library is and whether it is worth doing the switch to the library now.

mbarrien:
Made PR https://github.com/grafana/grizzly/pull/260 (removing adding Alert rules for now).

joanlopez:
Closing in favor of #262, let's try to get that one ready to merge! 🚀

PR #259, opened 2023-10-09
Deprecate --directory/-d flag in favor of os.Stat

joanlopez:
One of the few things that @undef1nd and I found a bit weird when starting to explore the Grizzly command-line interface is the need to specify the `--directory/-d` flag to say that the `<resource-path>` is a directory, an even more confusing that it's mandatory for the `grr pull` command.

So, this is both:
- A proposal to just get rid of this flag, and use `os.Stat` instead to determine whether the provided path `<resource-path>` is a directory or not
- A direct question to @malcolmholmes: Am I missing something? Was there any other reason/need to use that flag?

Note that the suggestion here is to keep it, but as deprecated, just to avoid breaking changes straight away.

Thanks!

### DIFF
--- a/pkg/grizzly/workflow.go
+++ b/pkg/grizzly/workflow.go
@@ -103,13 +103,19 @@ func ListRemote(opts Opts) error {
 	return w.Flush()
 }
 
-// Pulls remote resources
+// Pull pulls remote resources and stores them in the local file system.
+// The given resourcePath must be a directory, where all resources will be stored.
 func Pull(resourcePath string, opts Opts) error {
-	log.Infof("Pulling resources from %s", resourcePath)
-
-	if !(opts.Directory) {
-		return fmt.Errorf("pull only works with -d option")
+	stat, err := os.Stat(resourcePath)
+	if err != nil {
+		return err
 	}
+
+	if !stat.IsDir() {
+		return fmt.Errorf("pull <resource-path> must be a directory")
+	}
+
+	log.Infof("Pulling resources to %s", resourcePath)
 
 	for name, handler := range Registry.Handlers {
 		if !Registry.HandlerMatchesTarget(handler, opts.Targets) {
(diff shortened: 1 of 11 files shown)


pkg/grizzly/workflow.go
```diff
@@ -103,14 +103,20 @@ func ListRemote(opts Opts) error {
 	return w.Flush()
 }
 
-// Pulls remote resources
+// Pull pulls remote resources and stores them in the local file system.
+// The given resourcePath must be a directory, where all resources will be stored.
 func Pull(resourcePath string, opts Opts) error {
-	log.Infof("Pulling resources from %s", resourcePath)
+	stat, err := os.Stat(resourcePath)
+	if err != nil {
```
joanlopez, line 110: It's probably good idea to omit `ErrNotExist` errors here, cause in _Pull_ operations, the local directory might not exist, and can _just_ be created. Right? Wdyt? @malcolmholmes

malcolmholmes:
-d introduced a very different way of operating. Instead of parsing a single file, it processed a complete directory of files. This seemed like a big difference at the time justifying an argument to distinguish.

I'm more than happy though with your observation that the flag is superfluous. I guess I would keep -d, confirm that, in that case the arg is a dir, and allow the situation without an argument to also accept a dir.

pkg/grizzly/workflow.go
malcolmholmes: pull requires a dir, no? In which case, sure it can make the dir.

joanlopez: Yeah, agreed! I added a bunch of changes to handle that, as well as to test the expected behavior!

joanlopez:
> I'm more than happy though with your observation that the flag is superfluous. I guess I would keep -d, confirm that, in that case the arg is a dir, and allow the situation without an argument to also accept a dir.

I think that'd be the behavior with the current set of changes:
- It keeps the `-d`, just marked as deprecated (leaves a message when used).
- It checks that the `<resource-path>` is a directory, when needed, despite of the flag (which means that if you use the flag it's also checked).

Sounds good?

undef1nd (approved):
Nice! Thanks, Joan, for making this improvement.

PR #262, opened 2023-10-17
Switch to grafana-openapi-client-go

mbarrien:
Switches datasource, dashboard, and folder resources to using the grafana-openapi-client-go instead of accessing the REST API's directly. This should supersede #260 (and possibly #257 since it adds a field to provider for the OpenAPI client) 

There are some small issues:
* The OpenAPI models JSON rendering has a lot of `omitempty` annotations, which results in many fields no longer being present in the next pull after you change to this code (especially for datasources).
* Some tests not passing `make test`:
  * Some bugs in the OpenAPI spec
    * getDatasourceByName does not define a 404 response, so in Golang code had to fall back to runtime.APIError instead of an error specific to that function call.
    * postDashboard the `id` field is expected to be a string, but Grafana is return a number
  * Getting 409 and 412 errors on some tests which aren't expecting them.

Note that some of the functionality lost in #260 due to missing fields has been restored since grafana-openapi-client-go contains them all.

If you have previously run grizzly pull, then run grizzly pull again in the new code, you will get diffs like this on all previously pulled resources (anonymized from actual examples):

Datasources:
```
--- a/grafana/grizzly/datasources/datasource-Grafana Cloud k6.yaml
+++ b/grafana/grizzly/datasources/datasource-Grafana Cloud k6.yaml
@@ -4,11 +4,7 @@ metadata:
     name: Grafana Cloud k6
 spec:
     access: proxy
-    basicAuth: false
-    basicAuthUser: ""
-    database: ""
     id: 33
-    isDefault: false
     jsonData:
         backendUrl: https://k6.io
         environment: production
@@ -18,11 +14,8 @@ spec:
     name: Grafana Cloud k6
     orgId: 1
     readOnly: true
-    secureJsonFields: {}
     type: k6-datasource
     typeLogoUrl: /public/plugins/k6-app/img/logo.svg
     uid: Grafana Cloud k6
     url: proxy
-    user: ""
     version: 1
-    withCredentials: false
```

Dashboards:
```
--- a/grizzly/dashboards/example-folder/dashboard-bazqux.yaml
+++ b/grizzly/dashboards/example-folder/dashboard-bazqux.yaml
@@ -23,6 +23,7 @@ spec:
     editable: true
     fiscalYearStartMonth: 0
     graphTooltip: 0
+    id: 180
     links: []
     liveNow: false
     panels:
@@ -133,4 +134,5 @@ spec:
     timezone: ""
     title: Sample Dashboard
     uid: bazqux
+    version: 2
[...]

PR #265, opened 2023-11-13
Re-strip id and version fields

malcolmholmes:
After merging #262, we lost the code that removed the `id` and `version` fields. These, if present, cause issues in diffs and rename/uploads. They really are internal to Grafana and not needed in the presentation of a dashboard.


Duologic (approved):
As it is broken now, this can't get worse? Approving. (I need to ramp up my grizzly knowledge)

PR #270, opened 2023-11-23
Simplify handlers

malcolmholmes:
Each handler currently has two files (three including tests), one for the handler itself, and another for utility functions that are package global.

This is daft. Why have this separation when everything relates to the handler itself.

This PR merges all the logic relating to a handler into a single file. In some situations, this could itself lead to further code reduction.


julienduchesne (approved):
LGTM, assuming most of the moved code is the same. The extra level (functions outside of the handlers) didn't make much sense indeed when the handler has all it needs to do the API calls

PR #271, opened 2023-11-27
Add support for Contexts

malcolmholmes:
At present, Grizzly is configured through envvars. This is great for CI setups (e.g 12 factor apps), and perhaps if you only interact with a single backend. However, as soon as you have more than one backend, it becomes painful.

This PR adds support for "contexts", modelled after `kubectl`. Each context contains a full set of settings/credentials. Switching between them becomes trivial.


docs/content/authentication.md
```diff
@@ -5,47 +5,90 @@ title: "Authentication"
 
 # Authentication and Configuration
 
-Each system handles authentication differently. Authentication must be configured
-if resources are provided for a system.
+## Introducing Contexts
```
Duologic, line 8:
```suggestion
```
Introduction can simply be the first paragraph.

docs/content/authentication.md
```diff
@@ -5,47 +5,90 @@ title: "Authentication"
...
 ## Grafana Itself
-This tool interacts with Grafana via its REST API. For this, you will need to
-establish authentication credentials. These are provided to `grr` via
-environment variables.
 
-| Name | Description | Required | Default |
-| --- | --- | --- | --- |
-| `GRAFANA_URL` | Fully qualified domain name of your Grafana instance. | true | - |
-| `GRAFANA_USER` | Basic auth username if applicable. | false | `api_key` |
-| `GRAFANA_TOKEN` | Basic auth password or API token. | false | - |
```
Duologic, line 20: Is this completely gone or simply deprecated?

docs/content/authentication.md
```diff
@@ -5,47 +5,90 @@ title: "Authentication"
...
+Grizzly supports 'contexts' (much like `kubectl` contexts), allowing users to maintain
+configuration for multiple Grafana (and Mimir and Synthetic Monitoring) systems.
+
+By default, the `default` context is used. We will discuss how to use the default
+context first, then explain how to use multiple contexts thereafter.
+
+## Settings
+
+> **NOTE**: If you have used Grizzly previously with environment variables, you can
+  initialise the `default` context from your environment simply with `grr config init`.
```
Duologic, line 18: How about ` grr config init --from-env-variables`, assuming that this command would fail if the variables are not set?

pkg/config/config.go
```diff
@@ -0,0 +1,224 @@
+package config
```
Duologic, line 1: Lots of boiler plate in this file, perhaps there is a library out there already supporting this? 

https://github.com/spf13/viper maybe?

pkg/grafana/provider.go
```diff
@@ -1,41 +1,105 @@
...
+			return nil, fmt.Errorf("Error locating configuration file: %v", err)
+		}
+		if exists {
+			conf, err := config.Load()
+			if err != nil {
+				return nil, err
+			}
+			p.context = conf.Current()
+		} else {
+			conf, err := config.FromEnvironment()
```
Duologic, line 59: Do we want to show a deprecation notice here?

pkg/grafana/provider.go
```diff
@@ -1,41 +1,105 @@
...
+	_, err := p.Current()
+	if err != nil {
+		return nil, err
+	}
+	parsedUrl, err := url.Parse(p.context.Grafana.URL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid Grafana URL")
+	}
+
+	transportConfig := gclient.DefaultTransportConfig().WithHost(parsedUrl.Host).WithSchemes([]string{parsedUrl.Scheme})
```
Duologic, line 83: nit: A few linebreaks haven't hurt anyone.

Duologic (commented):
Great to see these different auth settings part of the same context, it provides a sense of order. The commands to configure a context is something that could be a page on Grafana Cloud, just copy paste and ready to go!

Perhaps also worth bringing this up with the auth team, this might be a feature they could own eventually.

docs/content/authentication.md
malcolmholmes: I have changed my mind since the code here. Now it will fall back to envvars with a warning. But there *are* usecases where envvars are totally appropriate (and should be documented). I will fix this.

docs/content/authentication.md
Duologic: `init` might sound more like something 'fresh' while falling back to env variables might be confusing.

pkg/config/config.go
malcolmholmes: Hmmm. I wonder about merging this and making that a separate issue/PR.

pkg/grafana/provider.go
malcolmholmes: Not any more. Looking at Viper, it places precedence of envvars *above* config. I've got it the other way around. I wonder if we should switch to match Viper's precedence, with the possibility of just adopting it.

pkg/grafana/provider.go
malcolmholmes: You clearly never coded in Java...

docs/content/authentication.md
```diff
@@ -5,13 +5,97 @@ title: "Authentication"
...
+Note, this will also work with other Cortex installations, alongside Grafana Cloud Prometheus/Mimir.
+
+## Grafana Synthetic Monitoring
+To interact with Grafana Synthetic Monitoring, you must configure the below settings:
+
+```sh
+grr config set synthetic-monitoring token abcdef123456 # API key (must have MetricsPublisher permissions)
+grr config set synthetic-monitoring stack-id # Grafana stack ID
+grr config set synthetic-monitoring metrics-id # Logs instance ID
+grr config set synthetic-monitoring logs-id # Metrics instance ID
```
julienduchesne, line 57:
```suggestion
grr config set synthetic-monitoring.token abcdef123456 # API key (must have MetricsPublisher permissions)
grr config set synthetic-monitoring.stack-id # Grafana stack ID
grr config set synthetic-monitoring.metrics-id # Logs instance ID
grr config set synthetic-monitoring.logs-id # Metrics instance ID
```

Missing some dots here I assume

docs/content/authentication.md
```diff
@@ -5,13 +5,97 @@ title: "Authentication"
...
+Grizzly contexts allow easy switching between multiple configurations and are best
+suited to workstation use.
+
+# Using Grizzly Contexts
+
+By default, the `default` context is used. We will discuss how to use the default
+context first, then explain how to use multiple contexts thereafter.
+
+> **NOTE**: If you have used Grizzly previously with environment variables, you can
+  initialise the `default` context from your environment simply with `grr config import`.
```
julienduchesne, line 23: So, Grizzly does not work without using that command first? Isn't that a breaking change?

pkg/grafana/dashboards_test.go
```diff
@@ -11,22 +11,12 @@ import (
 )
 
 func TestDashboard(t *testing.T) {
```
julienduchesne, line 13: Do we have any tests that run the actual grr commands? I feel like these tests are slightly short of testing the whole thing and it'd actually be more useful (full integration test) and easier to call the `grr` commands to do the work

pkg/config/config.go
```diff
@@ -0,0 +1,147 @@
...
+	viper.BindEnv("overrides.grafana.token", "GRAFANA_TOKEN")
+
+	viper.BindEnv("overrides.synthetic-monitoring.token", "GRAFANA_SM_TOKEN")
+	viper.BindEnv("overrides.synthetic-monitoring.stack-id", "GRAFANA_SM_STACK_ID")
+	viper.BindEnv("overrides.synthetic-monitoring.logs-id", "GRAFANA_SM_METRICS_ID")
+	viper.BindEnv("overrides.synthetic-monitoring.metrics-id", "GRAFANA_SM_LOGS_ID")
+
+	viper.BindEnv("overrides.mimir.address", "CORTEX_ADDRESS")
+	viper.BindEnv("overrides.mimir.tenant-id", "CORTEX_TENANT_ID")
+	viper.BindEnv("overrides.mimir.api-key", "CORTEX_API_KEY")
```
julienduchesne, line 35: For a next PR, we could support `MIMIR_` envvars and fallback to `CORTEX_` ones

docs/content/authentication.md
malcolmholmes: No, envvars override contexts. So this is a way to transfer your envvars. What we need to document is that you then need to *unset* your envvars otherwise they will continue to override.

pkg/config/config.go
malcolmholmes: Well, switch to calling out to mimirtool, which didn't exist when this integration was made. Then we could allow both CORTEX_ and MIMIR_ evvars, for back compat.

julienduchesne (approved):

PR #275, opened 2023-11-30
Update OpenAPI client to latest version

julienduchesne:
Gets rid of all the `params` objects
Also, there was an error on the folder update command it seems. The params didn't have the uid being set. Doesn't seem to be tested


julienduchesne:
Superseded by https://github.com/grafana/grizzly/pull/285

PR #282, opened 2023-12-14
feat: add support for library elements

theSuess:
This is a reimplementation/follow up to
https://github.com/grafana/grizzly/pull/218, using the openapi client to keep in line with the rest of the codebase. It allows pull & push of library panels and variables.

Tested pull & push with a simple library panel and a dashboard using this panel

### DIFF
--- a/pkg/grafana/library-elements-handler.go
+++ b/pkg/grafana/library-elements-handler.go
@@ -1,0 +1,219 @@
+package grafana
+
+import (
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+
+	"errors"
+
+	"github.com/grafana/grizzly/pkg/grizzly"
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
+
+	library "github.com/grafana/grafana-openapi-client-go/client/library_elements"
+	"github.com/grafana/grafana-openapi-client-go/models"
+)
+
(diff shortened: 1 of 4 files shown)


malcolmholmes:
Is there ordering for applies? E.g. folder-->panel-->dashboard? If not deterministic, we could get failures due to ordering.

theSuess:
Ah, I didn't catch that the handlers are ordered, should be in the right place now (after folders, before dashboards)

pkg/grafana/library-elements-handler.go
```diff
@@ -0,0 +1,225 @@
...
+	return h.updateElement(existing, resource)
+}
+
+func (h *LibraryElementHandler) listElements() ([]string, error) {
+	params := library.NewGetLibraryElementsParams()
+	client, err := h.Provider.(ClientProvider).Client()
+	if err != nil {
+		return nil, err
+	}
+	elemsOK, err := client.LibraryElements.GetLibraryElements(params, nil)
```
julienduchesne, line 149: You'll need to rebase following the OpenAPI client update. This function likely changed (no more params)

theSuess: done!

julienduchesne (approved):
LGTM. I tried it out and the issue raised in https://github.com/grafana/grizzly/pull/218 looks fixed. Let's :shipit: !

PR #285, opened 2023-12-15
chore: update openapi client

theSuess:
This updates the openapi client dependency to the latest version.

Requirement for supporting alerting since some options are only available in the latest version


theSuess:
Drone test seems to fail because of buildinfra issues. I don't have the permissions to restart the build though :/

julienduchesne (approved):
Heh, had the exact same PR up: https://github.com/grafana/grizzly/pull/275

PR #286, opened 2023-12-15
feat: add basic alerting resources

theSuess:
This allows for pull/push of alert rules & alert rule groups.

Will create a follow up PR for contact points & notification policies


theSuess:
Needs #285 first, otherwise created alerts will never be able to be removed

pkg/grafana/alertgroup-handler.go
```diff
@@ -0,0 +1,210 @@
...
+	return "json"
+}
+
+const (
+	alertRuleGroupGlob    = "alert-rules/alertRuleGroup-*"
+	alertRuleGroupPattern = "alert-rules/alertRuleGroup-%s.%s"
+)
+
+// FindResourceFiles identifies files within a directory that this handler can process
+func (h *AlertRuleGroupHandler) FindResourceFiles(dir string) ([]string, error) {
```
malcolmholmes, line 69: Is this a Grizzly concept, or something new? Same for next func.

pkg/grafana/alertrule-handler.go
```diff
@@ -0,0 +1,203 @@
...
+
+	"github.com/grafana/grizzly/pkg/grizzly"
+	"github.com/grafana/tanka/pkg/kubernetes/manifest"
+
+	"github.com/grafana/grafana-openapi-client-go/client/provisioning"
+	"github.com/grafana/grafana-openapi-client-go/models"
+)
+
+// AlertRuleHandler is a Grizzly Handler for Grafana alertRules
+type AlertRuleHandler struct {
```
malcolmholmes, line 18: Does a  AlerRuleHandler make sense? The API will just be updating a single rule inside an AlertRuleGroup right? So actually this makes things more complicated than they need be?

pkg/grafana/alertgroup-handler.go
theSuess: That's from grizzly. When performing the pull/push flow, grizzly stores resources based on the `ResourceFilePath` function and discovers them using `FindResourceFiles`

pkg/grafana/alertrule-handler.go
theSuess: I'm open to just having an `AlertRuleGroup` handler. It mostly comes down to ergonomics :shrug:

theSuess: I've pushed a PR now to reduce this to only use AlertRuleGroups

julienduchesne (approved):
Very nice. Tried all the commands. Everything works great! The code looks good to me. Let's :shipit:

NicoAdrian:
Any example how to use this new feature ?

PR #289, opened 2023-12-19
feat: add contact points

theSuess:
This adds contact points as a configurable resource

Complements #286


pkg/grafana/provider.go
```diff
@@ -79,5 +79,6 @@ func (p *Provider) GetHandlers() []grizzly.Handler {
 		NewDashboardHandler(p),
 		NewRuleHandler(p),
 		NewSyntheticMonitoringHandler(p),
+		NewContactPointHandler(p),
```
julienduchesne, line 82: Can we call it 
```suggestion
		NewAlertContactPointHandler(p),
```

So that it's grouped with https://github.com/grafana/grizzly/pull/286. Same for all other places where it's only `ContactPoint`. Maybe `AlertingContactPoint` and `AlertingRuleGroup` would be even better 🤔

julienduchesne:
Mostly works well but since most fields are secret and the API returns `[REDACTED]` for those, grizzly finds differences on every apply run

I don't know if that's OK

pkg/grafana/provider.go
theSuess: Renamed

theSuess:
> Mostly works well but since most fields are secret and the API returns `[REDACTED]` for those, grizzly finds differences on every apply run
> 
> I don't know if that's OK

We don't really have a choice here. If we were to ignore changes in the auth section, the user would be unable to update auth settings once set.

PR #290, opened 2023-12-20
feat: add notificationpolicy type

theSuess:
This adds the notification policy resource type. As the NotificationPolicy is a single global object, I've made the decision to just use "global" as the UID and validate this on parsing resources. It's also saved into the single "notificationpolicy.yaml" file directly.

AFAIK this is the first global resource we have but let me know if someone has a better idea on how to represent this.

(Also this updates the openapi client to include https://github.com/grafana/grafana-openapi-client-go/pull/50)


pkg/grafana/provider.go
```diff
@@ -79,5 +79,6 @@ func (p *Provider) GetHandlers() []grizzly.Handler {
 		NewDashboardHandler(p),
 		NewRuleHandler(p),
 		NewSyntheticMonitoringHandler(p),
+		NewNotificationPolicyHandler(p),
```
julienduchesne, line 82: Same comment as in https://github.com/grafana/grizzly/pull/289
I'd call it `AlertingNotificationPolicy`

julienduchesne (commented):
Works great as well! Only comment I have is about naming

pkg/grafana/provider.go
theSuess: I've renamed it to `AlertNotificationPolicy` to keep in line with `AlertRuleGroup` for now. If we want to rename this, we can do everything at once at a later point

julienduchesne (approved):
👍

PR #306, opened 2024-01-17
Allow non-glob and lower case targets

malcolmholmes:
The syntax for targets is currently:

    grr <cmd> -t Dashboard.*

This is powerful, in that it allows a user to pick out specific targets by UID
pattern.

However, I've never actually used, nor seen it used, for anything other than matching
by kind.

Therefore, this is simpler and likely to lead to less confusion:

    grr <cmd> -t Dashboard

For that matter, while the kind follows a capitalised pattern (as does Kubernetes),
it adds nothing here, other than risk for confusion. Thus this will now work too:

    grr <cmd> -t dashboard
﻿


Duologic (commented):
Does the repetition warrant an abstraction?

malcolmholmes:
> It does. Eventually.

Actually, there is only one duplication that was easily fixed.

Duologic (approved):

PR #310, opened 2024-01-19
Add reviewing proxy server to Grizzly

malcolmholmes:
This adds a VSCode Extension style proxy to Grizzly.

The [Grafana VSCode Extension](https://github.com/grafana/grafana-vs-code-extension) enables a user to
edit dashboards stored as JSON in their IDE using a Grafana UI. It does this by starting up an
embedded proxy server that intercepts requests for dashboard JSON, while forwarding other requests
to a backing Grafana instance.

The VSCode extension is made up of UI code alongside this proxy. The proxy is equally usable outside
of VSCode. This PR brings the proxy into Golang and makes it available to users outside of a specific
IDE.

Just using this proxy for a few minutes shows its potential within Grizzly. A user can:
```
grr config <... whatever is needed to set up your site....>
grr pull resources # pulls resources from remote Grafana
grr proxy -b resources # opens a proxy with all of the downloaded dashboards available for review using the proxy, and opens a browser
```


julienduchesne:
If there's no backing instance, and it tries to proxy requests, it currently panics. Perhaps we should catch these failures and show a better error message

cmd/grr/workflow.go
```diff
@@ -193,6 +193,26 @@ func previewCmd() *cli.Command {
 	return initialiseCmd(cmd, &opts)
 }
 
+func serveCmd() *cli.Command {
+	cmd := &cli.Command{
+		Use:   "serve <resources>",
+		Short: "Run Grizzly server",
+		Args:  cli.ArgsExact(1),
+	}
```
julienduchesne, line 201: This command needs a description and also, I'm not sure that `serve` is the correct command name for that. `serve-editor-proxy`, maybe?

malcolmholmes: It is likely to do more than just the proxy. E.g. things like `grr show` could appear in the UI, reviewing YAML/JSON, converting between, managing contexts, and more could happen via the UI as much as cli.

malcolmholmes:
> If there's no backing instance, and it tries to proxy requests, it currently panics. Perhaps we should catch these failures and show a better error message

This should be resolved.

cmd/grr/workflow.go
malcolmholmes: We'd like to merge then can discuss naming after, if needed.

K-Phoen (approved):

PR #312, opened 2024-01-24
Support nested folders

malcolmholmes:
Nested folders introduce a need for **ordering** in terms of when resources are pushed to Grafana.

This PR changes the the sort mechanisms to (a) sort by handler precedence and (b) allow a handler to
provide its own sorting mechanism. The `FolderHandler` is the only one, at present, that does anything
other than return the list of resources as is.


pkg/grafana/folder-handler.go
```diff
@@ -87,6 +89,34 @@ func (h *FolderHandler) GetUID(resource grizzly.Resource) (string, error) {
...
+	sorted := grizzly.Resources{}
+	for _, resource := range rootFolders {
+		s, _ := graph.TopSort(resource.UID())
+		for i := len(s) - 1; i >= 0; i-- {
+			sorted = append(sorted, foldersByUID[s[i]])
+		}
+	}
+	return sorted
+}
+
```
julienduchesne, line 119:
```suggestion

// Sort sorts according to handler needs
func (h *FolderHandler) Sort(resources grizzly.Resources) grizzly.Resources {
	result := grizzly.Resources{}
	addedToResult := map[string]bool{}
	for _, resource := range resources {
		addedToResult[resource.UID()] = false
	}
	for {
		continueLoop := false
		for _, resource := range resources {
			parentUID, hasParentUID := resource.Spec()["parentUid"]
			// Add root folders
			if !hasParentUID {
				addedToResult[resource.UID()] = true
				result = append(result, resource)
				continue
			}
			parentAdded, parentExists := addedToResult[parentUID.(string)]
			// Add folders with parents which aren't declared in Grizzly
			if !parentExists {
				addedToResult[resource.UID()] = true
				result = append(result, resource)
				continue
			}
			// Add folders with parents which have already been added
			if parentAdded {
				addedToResult[resource.UID()] = true
				result = append(result, resource)
				continue
			}
			// Delay folders with parents which haven't been added yet
			continueLoop = true
		}

		if !continueLoop {
			break
		}
	}

	return result
}
```

I feel like you don't need an external lib with possibly undefined behavior for this (ex: missing node). And you'll never have enough resources that an algorithm like this will be slow. This likely takes under a second for millions of folders

julienduchesne: Or with less lines:
```suggestion
// Sort sorts according to handler needs
func (h *FolderHandler) Sort(resources grizzly.Resources) grizzly.Resources {
	result := grizzly.Resources{}
	addedToResult := map[string]bool{}
	for _, resource := range resources {
		addedToResult[resource.UID()] = false
	}
	for {
		continueLoop := false
		for _, resource := range resources {
			parentUID, hasParentUID := resource.Spec()["parentUid"]
			parentAdded, parentExists := addedToResult[parentUID.(string)]
			// Add
			// - root folders
			// - folders with parents which aren't declared in Grizzly
			// - folders with parents which have already been added
			if !hasParentUID || !parentExists || parentAdded {
				addedToResult[resource.UID()] = true
				result = append(result, resource)
				continue
			}
			// Delay folders with parents which haven't been added yet
			continueLoop = true
		}

		if !continueLoop {
			break
		}
	}

	return result
}
```

malcolmholmes: Thx, will review.

malcolmholmes: @julienduchesne I committed your proposal, with one addition: you need to check for values already added to the addedToResult map and skip them.

pkg/grafana/folder-handler.go
```diff
@@ -87,6 +88,58 @@ func (h *FolderHandler) GetUID(resource grizzly.Resource) (string, error) {
 	return resource.Name(), nil
 }
 
+type folderForSorting struct {
```
julienduchesne, line 91: This type is unused

pkg/grafana/folder-handler.go
```diff
@@ -87,6 +88,58 @@ func (h *FolderHandler) GetUID(resource grizzly.Resource) (string, error) {
...
+	UID      string
+	Resource grizzly.Resource
+}
+
+// Sort sorts according to handler needs
+func (h *FolderHandler) Sort(resources grizzly.Resources) grizzly.Resources {
```
julienduchesne, line 97: Would be pretty easy to write a unit test for this. I can do it if you don't want to spend the time 😄

malcolmholmes: It would be too easy to leave it to you. Not good for my character. I'll take a go!

joschi36:
Just were looking for this and just tried it. However, folders always just created in the root level. Sorry if its not finished yet or I missed something. As I was just looking for a solution and found this, I was happy to quickly give it a go.

pkg/grafana/folder-handler.go
julienduchesne: Done!

malcolmholmes:
@joschi36 I found that too and fixed it. But, it looks like I might not have committed it before heading out. I'll see if I can get it pushed here.

malcolmholmes:
@joschi36 I see the code is there, so I did push it. Are you seeing the issue with initial creation, or updating folders?

pkg/grafana/folder-handler.go
malcolmholmes: removed

julienduchesne (approved):
LGTM! 👍

PR #313, opened 2024-01-26
Accept files anywhere within resource path

malcolmholmes:
Grizzly needs to know a specific file structure when writing files. E.g. do dashboards
go into a `dashboards/` folder?

The way Grizzly is currently coded, Grizzly also expects files to follow this format
when being read. There's no good reason for this, given that all files are a "resource"
and Grizzly will pass them to the relevant handler based upon their apiVersion and kind.﻿


julienduchesne (approved):
Awesome! I was wondering about that the other day

malcolmholmes:
Replaced by #336

Issue #318, opened 2024-01-31
Remote communication broken since v0.3.0 when using Grafana on non-root path


sberlin:
Since v0.3.0 Grizzly is broken for us. We receive differing error messages when using commands like pull or apply. A selection of those errors:
```
/src # grr pull -d foo
Flag --directory has been deprecated, now it is inferred from the operating system
INFO[0000] Pulling resources to foo
FATA[0000] [GET /v1/provisioning/alert-rules] GetAlertRules (status 404): {}
/src # grr pull -d foo
Flag --directory has been deprecated, now it is inferred from the operating system
INFO[0000] Pulling resources to foo
FATA[0000] [GET /v1/provisioning/alert-rules] GetAlertRules (status 404): {}
/src # grr pull -d foo
Flag --directory has been deprecated, now it is inferred from the operating system
INFO[0000] Pulling resources to foo
FATA[0000] [GET /v1/provisioning/alert-rules] GetAlertRules (status 404): {}
FATA[0000] &{ <nil> } (*models.ErrorResponseBody) is not supported by the TextConsumer, can be resolved by supporting TextUnmarshaler interface
```

Connection is setup with
```
/src # export GRAFANA_URL=http://localhost:8080/grafana
/src # export GRAFANA_TOKEN=foo
/src # grr config import # optionally
```

When running a webserver at that address, I receive the following
```
docker run --rm -p 8080:8080 -v "$PWD:/app:z" -u "$(id -u):0" bitnami/nginx:latest
[...]
2024/01/31 15:16:32 [error] 39#39: *11 open() "/opt/bitnami/nginx/html/api/search" failed (2: No such file or directory), client: 10.0.2.100, server: , request: "GET /api/search?limit=1000&page=1&type=dash-folder HTTP/1.1", host: "192.168.178.20:8080"
10.0.2.100 - - [31/Jan/2024:15:16:32 +0000] "GET /api/search?limit=1000&page=1&type=dash-folder HTTP/1.1" 404 117 "-" "Go-http-client/1.1" "-"
2024/01/31 15:19:57 [error] 39#39: *12 open() "/opt/bitnami/nginx/html/api/v1/provisioning/alert-rules" failed (13: Permission denied), client: 10.0.2.100, server: , request: "GET /api/v1/provisioning/alert-rules HTTP/1.1", host: "192.168.178.20:8080"
[...]

PR #320, opened 2024-02-01
Use docker compose to run the integration instance

julienduchesne:
This removes the need for a docker build and it will allow me to run a second instance for https://github.com/grafana/grizzly/issues/318 
The second one will have a sub-path configured

Other docker compose advantages:
 - Allows to turn down the stack after tests
 - Will recreate the stack from scratch each time


malcolmholmes (approved):
Note, this instance isn't just used by integration tests. It is also used by the main tests (they just invoke Handlers directly, so not fully integration tests).

julienduchesne:
> Note, this instance isn't just used by integration tests. It is also used by the main tests (they just invoke Handlers directly, so not fully integration tests).

Should I put it closer to the root path. `./test-docker-compose` or something

PR #321, opened 2024-02-01
Add new context and make integration tests stricter

julienduchesne:
I added a new context and I noticed that the tests were still passing (incorrectly)
Turns out the `.Contains` condition on the output was still matching. I changed to a strict `Equal` and removed the stderr output from there. We're not trying to match log lines here

The subpath context is unused so far, but it will be in a future PR

### DIFF
--- a/integration/utils_test.go
+++ b/integration/utils_test.go
@@ -1,7 +1,6 @@
 package integration_test
 
 import (
-	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


test-docker-compose/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '3'
```
malcolmholmes, line 1: Could these files be inside `integration`?

integration/utils_test.go
```diff
@@ -33,14 +32,14 @@ func RunTests(t *testing.T, tests []GrizzlyTest) {
...
 				cmd.Dir = test.TestDir
-				output, err := cmd.CombinedOutput()
+				output, err := cmd.Output()
 				if command.ExpectedError != nil {
 					require.Error(t, err, command.ExpectedError)
 				}
 				if command.ExpectedOutput != "" {
 					data, err := os.ReadFile(filepath.Join(test.TestDir, command.ExpectedOutput))
-					require.NoError(t, err, fmt.Sprintf(""))
-					require.Contains(t, string(output), string(data))
```
malcolmholmes, line 43: we could add `command.ExpectedContains` if we want the `Contains` functionality back.

malcolmholmes (approved):

integration/utils_test.go
julienduchesne: Yeah, it would be unused for now. We can add it when we need it, right?

test-docker-compose/docker-compose.yml
julienduchesne: What do you mean? The `docker-compose` config is used for both integration tests and normal tests

integration/utils_test.go
malcolmholmes: exactly

test-docker-compose/docker-compose.yml
malcolmholmes: Okay.

I wonder if we could port the normal test functionality into integration tests, at least for those that depend upon a Grafana instance, as otherwise they're not really unit tests.

julienduchesne: Agreed. Integration tests are all we need, IMO. It tests the functionality end-to-end

julienduchesne: Let's move the whole thing back to the integration dir, when and if we start only using these instances there. Does that work?

malcolmholmes: sure!

PR #323, opened 2024-02-01
Integration: Add utility to run the same test on multiple contexts

julienduchesne:
Closes https://github.com/grafana/grizzly/issues/26

To verify issues like above, we need to be able to test with various auth methods 
This PR adds a `basic_auth` context (and instance) and adds the functionality to run the same test on all contexts. Running the same tests lots of times on various set-ups is also a good way to find flaky features

Also, I changed from the `RunTests` function to individual t.Run. This is helpful because vscode can now see the individual tests and we can run them individually

See `run test | debug test` buttons in the follow screenshot
image


malcolmholmes (approved):
Looks good to me!

Issue #315, opened 2024-01-30
Adding contexts


malcolmholmes:
"Contexts" are a major improvement to Grizzly's usability. Now, instead of having to set (and reset) environment variables, they can be set once, and users can easily switch between them.

As well as storing credentials, contexts can also store other settings, such as:
* `only-spec`: Tells Grizzly only to store the raw file content without the envelope
* `output-format`: Allows switching between YAML and JSON when writing files

This issue is intended to group together these tasks and usability improvements.

PR #329, opened 2024-02-02
Add config "get" and "path" subcommands

julienduchesne:
Related to https://github.com/grafana/grizzly/issues/315

- get: Allows quickly getting the current config without going into the YAML settings file
- path: Allows actually finding the current config file. Had trouble finding the `/Users/julienduchesne/Library/Application Support/grizzly/settings.yaml` on my machine. This helps

While implementing the `config path` function, I noticed that the config logic was a bit weird. While grizzly was possibly reading (through viper) the config file from the workdir, it was writing back to the system config. Is this expected? By using `viper.ConfigFileUsed()`, we write back to the file we're reading the config from

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,14 +1,15 @@
 package config
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
-	"path/filepath"
 	"sort"
 	"strings"
 
 	"github.com/kirsle/configdir"
 	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


pkg/config/config.go
```diff
@@ -1,14 +1,15 @@
...
+	"encoding/json"
 	"fmt"
 	"os"
-	"path/filepath"
 	"sort"
 	"strings"
 
 	"github.com/kirsle/configdir"
 	"github.com/spf13/viper"
+	"gopkg.in/yaml.v3"
```
malcolmholmes, line 12: We use github.com/goccy/go-yaml not this one in Grizzly.

malcolmholmes (approved):

pkg/config/config.go
julienduchesne: hmm, it's not in the go.mod file

malcolmholmes: Ahh right, that's in an unmerged branch. Ignore this comment then.

Issue #216, opened 2022-06-12
Synthetic Monitoring region support


dezhin:
It seems Grizzly supports only US region and endpoint address is hard-coded [here](https://github.com/grafana/grizzly/blob/fc8bdd017774463b02125dda677e9a425bf875d8/pkg/grafana/synthetic-monitoring.go#L45).

That's why SM checks that I create with Grizzly aren't visible in the web interface, but can be pulled with Grizzly. Moreover Grizzly cannot pull previously created checks from SM/

spinillos:
Region support was fixed in https://github.com/grafana/grizzly/pull/420.

PR #330, opened 2024-02-02
Config: Allow configuring SM URL

julienduchesne:
Closes https://github.com/grafana/grizzly/issues/216

Also, fix the `get` command so that it returns the real config. Not just what's in the config file but also default values and environment var values

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -30,6 +30,7 @@ func override(v *viper.Viper) {
 		"grafana.user":  "GRAFANA_USER",
 		"grafana.token": "GRAFANA_TOKEN",
 
+		"synthetic-monitoring.api-url":    "GRAFANA_SM_API_URL",
 		"synthetic-monitoring.token":      "GRAFANA_SM_TOKEN",
 		"synthetic-monitoring.stack-id":   "GRAFANA_SM_STACK_ID",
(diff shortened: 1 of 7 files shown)


pkg/config/config.go
```diff
@@ -121,20 +122,31 @@ func UseContext(context string) error {
...
 	}
 	contextPath := fmt.Sprintf("contexts.%s", name)
 	ctx := viper.Sub(contextPath)
 	if ctx == nil {
 		ctx = viper.New()
 	}
+
+	// Set defaults
+	if !ctx.IsSet("synthetic-monitoring.api-url") {
+		ctx.Set("synthetic-monitoring.api-url", "https://synthetic-monitoring-api.grafana.net")
```
malcolmholmes, line 139: Hardwired value in config? If hardwired, surely that would be in the handler?

julienduchesne: I don't know. It's a default value to a configuration, not really hardwired. It's more visible this way, and that's why it's more easily testable.

Either way works, I can switch to figuring out the URL in code 🤔

malcolmholmes: `viper` has `viper.SetDefault("ContentDir", "content")` that would be more appropriate here, and could happen earlier? I guess upon creation of a new context? I wonder if we could make defaults a per-handler thing, where we defer to each handler to configure its defaults. Then, for Grafana we could default to http://localhost:3000/admin/admin for example?

PR #333, opened 2024-02-05
Clearer dashboard validation

malcolmholmes:
We now have `Validate()` and `Prepare()`. Validate, well, validates. That is, it either returns nil
or an error. Prepare mutates the resource.

This PR moves these two related funcs closer to each other in the code. Also, for dashboards it
refactors the validation logic:

* if UID not set or ="", then use name as UID
* if UID and name set, but not equal, throw an error
* if name is blank, throw an error.﻿

### DIFF
--- a/pkg/grafana/dashboard-handler.go
+++ b/pkg/grafana/dashboard-handler.go
@@ -89,11 +89,11 @@ func (h *DashboardHandler) Prepare(existing, resource grizzly.Resource) *grizzly
 // Validate returns the uid of resource
 func (h *DashboardHandler) Validate(resource grizzly.Resource) error {
 	uid, exist := resource.GetSpecString("uid")
+	if resource.Name() == "" {
+		return fmt.Errorf("name cannot be blank")
+	}
 	if resource.Name() != uid && exist {
 		return fmt.Errorf("uid '%s' and name '%s', don't match", uid, resource.Name())
-	}
-	if resource.Name() == "" && (!exist || uid == "") {
-		return fmt.Errorf("uid and name cannot be blank")
 	}
 	return nil
 }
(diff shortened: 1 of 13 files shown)


pkg/grafana/dashboard-handler.go
```diff
@@ -90,9 +79,25 @@ func (h *DashboardHandler) Unprepare(resource grizzly.Resource) *grizzly.Resourc
...
+	}
 	return &resource
 }
 
+// Validate returns the uid of resource
+func (h *DashboardHandler) Validate(resource grizzly.Resource) error {
+	uid, exist := resource.GetSpecString("uid")
+	if resource.Name() == "" {
+		return fmt.Errorf("name cannot be blank")
+	}
```
julienduchesne, line 94: Isn't that a base condition for all resources?

pkg/grafana/dashboard_test.go
```diff
@@ -81,4 +81,59 @@ func TestDashboard(t *testing.T) {
...
+						require.NoError(t, err)
+					} else {
+						require.Contains(t, err.Error(), test.ValidateErrorString)
+					}
+				} else {
+					require.NoError(t, err)
+					newResource := handler.Prepare(nil, test.Resource)
+					uid, _ := handler.GetUID(*newResource)
+					require.Equal(t, uid, test.ExpectedUID)
+				}
```
julienduchesne, line 135:
```suggestion
				err := handler.Validate(test.Resource)
if test.ValidateErrorString != "" {
	require.Contains(t, err.Error(), test.ValidateErrorString)
	return
}

require.NoError(t, err)
newResource := handler.Prepare(nil, test.Resource)
uid, _ := handler.GetUID(*newResource)
require.Equal(t, uid, test.ExpectedUID)
```

I think you can remove a level of `if` here

julienduchesne (approved):

pkg/grafana/dashboard-handler.go
malcolmholmes: Yeah. I did wonder while typing it. Hmm.

pkg/grafana/dashboard_test.go
malcolmholmes: thx. Makes sense.

pkg/grafana/dashboard-handler.go
malcolmholmes: I've updated the PR to do some validation for every resource before calling `handler.Validate()`.

julienduchesne (approved):

PR #334, opened 2024-02-07
Add --kind to set the kind of object when --only-spec is set

malcolmholmes:
The `--only-spec` argument allows for working with resources that lack an "envelope". This envelope
provides a way to specify metadata, e.g. resource type (aka 'kind'), plus in the case of dashboards,
the UID of the folder the dashboard belongs in.

If there is no envelope, then this information needs to be provided in another way. Currently, it is simply hardwired to dashboards. However, this PR changes this such that it can either be set in the context, or can be set on the command line each time Grizzly is invoked.﻿

### DIFF
--- a/pkg/grizzly/workflow.go
+++ b/pkg/grizzly/workflow.go
@@ -543,16 +543,12 @@ func getOnlySpec(opts Opts) (bool, string, string, error) {
 	if !onlySpec {
 		return false, "", "", nil
 	}
-	var kind string
-	if context.ResourceKind != "" {
-		kind = context.ResourceKind
-	} else {
+	kind := context.ResourceKind
+	if kind == "" {
 		kind = opts.ResourceKind
 	}
-	var folderUID string
-	if context.FolderUID != "" {
-		folderUID = context.FolderUID
-	} else {
+	folderUID := context.FolderUID
+	if folderUID == "" {
 		folderUID = opts.FolderUID
 	}
 	return onlySpec, kind, folderUID, nil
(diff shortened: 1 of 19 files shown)


pkg/grizzly/workflow.go
```diff
@@ -528,3 +528,33 @@ func getOutputFormat(opts Opts) (string, bool, error) {
...
+		kind = opts.ResourceKind
+	}
+	var folderUID string
+	if context.FolderUID != "" {
+		folderUID = context.FolderUID
+	} else {
+		folderUID = opts.FolderUID
+	}
+	return onlySpec, kind, folderUID, nil
+}
```
malcolmholmes, line 560: This whole function is ugly. We have a way for envvars to override context values (partially provided by Viper). However, we have not yet established a meaningful (and simple) way for CLI arguments to override context values. Therefore, we require a hack like this to conditionally extract values from either the context or the CLI depending. There's going to be a better way to do this, but that's for a different PR.

Duologic: But yeah, another PR sounds reasonable.

spinillos: You can reduce a bit the code if you change:

```
var kind string
if context.ResourceKind != "" {
  kind = context.ResourceKind
} else {
  kind = opts.ResourceKind
}
```

with this:
```
kind := context.ResourceKind
if kind == "" {
  kind = opts.ResourceKind
}
```

And same with folderUID.

Other idea could be remove the bool return and return a typed error in [this](https://github.com/grafana/grizzly/pull/334/files#diff-1c2bc400869dd54ad3f708aebc293bb446249797cd4d64de029df22eaf5376fcR545) line.

Duologic (approved):
LGTM

PR #335, opened 2024-02-07
Utility to generate jsonnet imports

malcolmholmes:
Grizzly already has support for the use of Jsonnet to generate resources. It also has
the ability to pull resources from Grafana/etc. However, it lacks the ability to
mutate those downloaded resources with Jsonnet.

This is due to Jsonnet's lack of dynamic imports. Due to this lack, it is not possible
to write a Jsonnet program that loads a set of files from within a directory, if the
filenames are not known at 'compile' time.

This is worked around by generating a file that contains a set of import statements that
can be consumed by a Jsonnet script.

This PR adds a command to generate such an imports file. The output will be of this
command will be of the form shown below. Jsonnet scripts can now iterate over the objects
imported in this file:

```
{
  "dashboard-id-test": {
    "path": "odoko/dashboards/odoko/dashboard-id-test.json",
    "type": "json",
    "resource": import "odoko/dashboards/odoko/dashboard-id-test.json",
  },
  "test-dashboard": {
    "path": "odoko/dashboards/test-dashboard.json",
    "type": "json",
    "resource": import "odoko/dashboards/test-dashboard.json",
  },
}
```

This would be generated with a command such as:

```
grr jsonnet imports dir imports.libsonnet
```

### DIFF
--- a/cmd/grr/config.go
+++ b/cmd/grr/config.go
@@ -9,6 +9,23 @@ import (
 	"github.com/spf13/viper"
 )
 
+func configCmd() *cli.Command {
+	cmd := &cli.Command{
+		Use:   "config <sub-command>",
+		Short: "Show, select or configure configuration",
+		Args:  cli.ArgsExact(0),
+	}
+	cmd.AddCommand(configPathCmd())
+	cmd.AddCommand(currentContextCmd())
+	cmd.AddCommand(useContextCmd())
+	cmd.AddCommand(getContextsCmd())
+	cmd.AddCommand(configImportCmd())
+	cmd.AddCommand(getConfigCmd())
+	cmd.AddCommand(setCmd())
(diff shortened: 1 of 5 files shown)


julienduchesne:
I don't think this should go in Grizzly tbh. Adding too many commands just leads to confusion, and this command isn't really "integrated" with other commands. This can be done with a simple bash script

julienduchesne (commented):
I don't think this should go in Grizzly tbh. Adding too many commands just leads to confusion, and this command isn't really "integrated" with other commands. This can be done with a simple bash script

malcolmholmes:
Fair enough. It isn't seeing traction, and is becoming stale. I'm not currently planning on merging it.

malcolmholmes:
Closing as original demand did not materialise

PR #336, opened 2024-02-08
Read files anywhere, not in fixed directory structure

malcolmholmes:
When Grizzly pulls resources from Grafana, it writes them in a pre-defined file system structure.

Currently, when reading files, Grizzly expects them in this same structure. This is unnecessary.

This PR changes Grizzly to accept files anywhere within the referenced directory.﻿

### DIFF
--- a/pkg/grafana/alertgroup-handler.go
+++ b/pkg/grafana/alertgroup-handler.go
@@ -2,7 +2,6 @@ package grafana
 
 import (
 	"fmt"
-	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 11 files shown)


julienduchesne (approved):
:shipit:

julienduchesne:
Next step could be allowing to define the pull pattern with gotemplate. That way users can pull to a flat (`{{ .Kind }}-{{ .Name }}.yml` structure if they wish. That's what I'd choose as a pattern, personally

PR #339, opened 2024-02-16
Correctly deduce UID, and track isDir

malcolmholmes:
When a resource (likely a dashboard) is provided in `only-spec` format, then Grizzly needs a way to deduce
the UID. This provides that functionality.

It also adds the ability for mutation of `Opts`, which allows parsing to identify whether a file or a directory
was parsed. This will be useful later.﻿

### DIFF
--- a/pkg/grafana/library-element-handler.go
+++ b/pkg/grafana/library-element-handler.go
@@ -100,6 +100,14 @@ func (h *LibraryElementHandler) GetUID(resource grizzly.Resource) (string, error
 	return resource.Name(), nil
 }
 
+func (h *LibraryElementHandler) GetSpecUID(resource grizzly.Resource) (string, error) {
+	spec := resource["spec"].(map[string]interface{})
+	if val, ok := spec["uid"]; ok {
+		return val.(string), nil
+	}
+	return "", fmt.Errorf("UID not specified")
+}
+
 // Sort sorts according to handler needs
 func (h *LibraryElementHandler) Sort(resources grizzly.Resources) grizzly.Resources {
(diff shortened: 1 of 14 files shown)


pkg/grafana/notificationpolicy-handler.go
```diff
@@ -79,6 +79,14 @@ func (h *AlertNotificationPolicyHandler) GetUID(resource grizzly.Resource) (stri
 	return resource.Name(), nil
 }
 
+func (h *AlertNotificationPolicyHandler) GetSpecUID(resource grizzly.Resource) (string, error) {
+	spec := resource["spec"].(map[string]interface{})
+	if val, ok := spec["XXXXXXX"]; ok {
```
julienduchesne, line 84: Should probably throw a non implemented error right off the bat for there

pkg/grafana/library-element-handler.go
```diff
@@ -100,6 +100,14 @@ func (h *LibraryElementHandler) GetUID(resource grizzly.Resource) (string, error
 	return resource.Name(), nil
 }
 
+func (h *LibraryElementHandler) GetSpecUID(resource grizzly.Resource) (string, error) {
+	spec := resource["spec"].(map[string]interface{})
+	if val, ok := spec["uid"]; ok {
+		return val.(string), nil
+	}
+	return "", fmt.Errorf("UID not specified")
```
julienduchesne, line 108: All of these can be refactored to a helper function
```suggestion
         return specAttribute("uid")
```

julienduchesne (approved):

PR #344, opened 2024-02-22
Move integration tests to integration tests dir

malcolmholmes:
Currently, most of the unit tests are actually integration tests, relying upon a running
Grafana instance. Whilst the majority of tests have relied upon this instance, this has
not been a problem.

This PR moves the tests that require this instance to the `/integration` directory, and
removes the setup of the instance for the unit tests.

This also fixes an erroneous addition to the test dashboard in the examples directory.﻿


malcolmholmes:
> Nice! Without the docker compose set-up, the tests outside of the integration dir will truly be unit tests

yes, they now run in milliseconds rather than tens of seconds, which is how they should be.

PR #346, opened 2024-02-22
Consistent parsing for all types

malcolmholmes:
Currently, Grizzly can parse single resources as JSON/YAML, but will crawl a tree for enveloped
resources when they are generated with Jsonnet. This gives Jsonnet an undue advantage. Grizzly
uses Tanka functionality to seek out these Kubernetes objects.

Therefore, this PR pulls the "Walk" functionality out of Tanka and into Grizzly, and makes it
work for itself. Now, a file when parsed will either be enveloped, not enveloped or crawled,
in that order.

### DIFF
--- a/pkg/grizzly/parsing.go
+++ b/pkg/grizzly/parsing.go
@@ -9,13 +9,12 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/google/go-jsonnet"
 	"github.com/grafana/grizzly/pkg/config"
 	"github.com/grafana/tanka/pkg/jsonnet/native"
-	"github.com/grafana/tanka/pkg/kubernetes/manifest"
-	"github.com/grafana/tanka/pkg/process"
 	log "github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
 )
@@ -31,7 +30,7 @@ func Parse(registry Registry, resourcePath string, opts *Opts) (Resources, error
 	}
 	opts.IsDir = true
 
-	var resources Resources
+	var parsedResources Resources
 	files, err := FindResourceFiles(resourcePath)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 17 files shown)


pkg/grizzly/parsing.go
```diff
@@ -279,70 +283,135 @@ func ValidateEnvelope(data map[string]any) error {
...
+		return err
 	}
+	return nil
+}
 
-	result, err := vm.EvaluateSnippet(jsonnetFile, script)
-	if err != nil {
-		return nil, err
+// walkJSON recurses into either a map or list, returning a list of all objects that look
+// like kubernetes resources. We support resources at an arbitrary level of nesting, and
```
julienduchesne, line 300: kubernetes?

pkg/grizzly/parsing.go
```diff
@@ -9,13 +9,12 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/google/go-jsonnet"
 	"github.com/grafana/grizzly/pkg/config"
 	"github.com/grafana/tanka/pkg/jsonnet/native"
```
julienduchesne, line 17: After this PR, we should just remove tanka completely as a dependency. Native functions don't do much anymore, all of their features have been introduced into the stdlib

julienduchesne (approved):
Kinda big, I might have missed somethings but it LGTM

I would not talk about `Kubernetes` in code comments, it's confusing. We're dealing with manifests here but they're not Kubernetes manifests

pkg/grizzly/parsing.go
malcolmholmes: Ahh, thanks for that. I tried removing Tanka then hit the nativeFuncs issue. I will look at removing Tanka properly again.

malcolmholmes:
> Kinda big, I might have missed somethings but it LGTM
> 
> I would not talk about `Kubernetes` in code comments, it's confusing. We're dealing with manifests here but they're not Kubernetes manifests

Yeah. Where I noticed k8s references, I removed them. I clearly missed some. Will fix.

PR #347, opened 2024-02-23
Don't propagate opts outside cmd

malcolmholmes:
Builds on #346.

Currently, the CLI parses flags into opts, then hands these down into the core workflows. Then, the
core workflows need to decide whether they should be using flags, or the viper based context configs.

Thus, workflows are having to make decisions that should already have been made before the workflows
were invoked (i.e. whether a specific value should use a flag, envvar or config value).

Whilst there's more to do, this PR simply prevents `opts` from leaving the `cmd/grr` package. Values
that workflows need are explicitly passed into those workflow functions. The intention of this PR is
to leave any bugs/inconsistencies in place (e.g. jsonnetPaths can only be set with a flag, not in
context).

A subsequent PR can put in place a more workable flag/envvar/config setup.


malcolmholmes:
> hmmm 🤔. I liked the old way better personally. The code seems messier now. Maybe the options shouldn't be parsed past the `cmd` package, but the Options struct could be in a common config package, and we could keep passing a struct through

Yeah, where a struct makes sense, we can. The principal aim of this PR is to make communication between cmd and workflow more explicit. Then, we can make resolving of flags/config/envvars have happened already before values get to the workflows.

So this PR is a step on a journey rather than a final resting place.

julienduchesne (approved):
LGTM 🤷 Can't say I looked at everything, it's a pretty big PR

PR #348, opened 2024-02-23
Remove Tanka dependency

malcolmholmes:
A previous PR removed Grizzly's dependency on Tanka for walking a tree seeking out
resources. The only remaining Tanka dependency is native functions. Most are now in
the standard lib, so native funcs are no longer required. Three are still needed, so
this PR adds them directly to Grizzly.


julienduchesne (approved):

pkg/grizzly/jsonnet.go
```diff
@@ -1,9 +1,13 @@
 package grizzly
 
 import (
+	_ "embed"
```
K-Phoen, line 4: What is that useful for? (A comment explaining this import would be useful)

K-Phoen (approved):

PR #349, opened 2024-02-23
Remove 'preview' functionality

malcolmholmes:
It was obvious long ago that Grizzly could be useful in the "review" phase of working
with dashboards as code. (That is, first we create, then we review, then we publish.)

First, Grizzly implemented preview via snapshots. Firstly, this only works for dashboards.
Secondly, snapshots are not an often used feature, and snapshot the data as well as the
dashboard. The user experience was not captivating.

Next came `Listen`. This listened on a websocket for changes in Grafana. With this, a
dashboard changed in Grafana could be saved to local disk within the listening Grizzly.
However, this was based upon a websocket feature in Grafana that had an API that wasn't
considered stable, so it was removed.

Now, Grizzly has a `serve` command. Its principal function at present is to allow the
review of generated dashboards, and editing of static json dashboards. It solves the
problem that both preview and listen were attempting to solve.

This PR, therefore, removes the preview functionality that I have not heard of in use
for years.

It also removes a left over `Listen` method, for which the CLI connection was removed a long time ago.


K-Phoen (approved):

mikaello:
Just to chime in as a user: We have been quite heavy users of `grr preview` by multiple teams in my org.

We have developed a grafana-jsonnet pipeline (GitLab) that users could include in their jsonnet-projects to get previews in every pull request they created in their projects. The snapshots are uploaded to a test instance with the same datasources as in prod.

malcolmholmes:
We've discussed this. Current thought: calling this feature "preview" is confusing because it has side-effects and preview doesn't imply that.

What if we reverted the removal and renamed the features `snapshot`?

malcolmholmes:
Actually, please comment on #389 .

PR #350, opened 2024-02-23
Establish multiple providers

malcolmholmes:
Grizzly currently has three "endpoints" in one provider: Grafana, Cortex/Mimir and Synthetic Monitoring. Each of these 
has its own authentication and configuration. Really, they should be separate providers.

This PR resolves that.

Also, whilst in previous PRs we have removed some use of globals, still various places can call `config.CurrentContext()`
to pull in config, anywhere in the code.

This PR prevents this happening. It makes sure that providers have their configuration injected in at creation so that they 
don't need to pull it in via global functions.


julienduchesne (approved):
👍

PR #353, opened 2024-02-28
Reapplying dashboards fails

malcolmholmes:
It appears in our transition to the new openapi client, we failed to test re-apply of
resources (notably dashboards). This is a significant regression.﻿

### DIFF
--- a/integration/dashboard_test.go
+++ b/integration/dashboard_test.go
@@ -60,6 +60,22 @@ func TestDashboard(t *testing.T) {
 		})
 	})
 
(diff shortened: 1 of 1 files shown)


malcolmholmes:
Turns out, this is either a false alarm, or has been fixed between 0.3.1 and `main`.

malcolmholmes:
But tests are worth having all the same.

julienduchesne (approved):

K-Phoen (approved):

PR #359, opened 2024-02-29
Make the port configurable in `grr serve`

K-Phoen:
The port `8080` is commonly used, so let's make sure that people aren't prevented from using the server because of a port conflict.


K-Phoen:
I didn't find anything that set this value, no. Even via viper :shrug: (context file or environment variables)

PR #361, opened 2024-02-29
First pass at 'continue on error'

malcolmholmes:
Currently, Grizzly fails on first error. This makes sense in scenarios where errors are not
considered acceptable, but when a single resource failing to upload is acceptable, Grizzly's
fail fast behaviour becomes problematic.

This PR adds a switch (`-e`) to `grr apply` that tells it to continue on beyond any errors.
Errors are collected, then returned as an aggregate to the user at the end of the process.﻿

This PR now does a few other things:
* Removes use of `log` to communicate with users, uses its own `notifier`
* Switches `Parse()` to return a single resource - no resource types need to return multiple
* Adds a simple "%d errors occurred, %d resources applied" summary at the end of a run.


malcolmholmes:
This is an untested first pass...

malcolmholmes:
Replaced by #371

PR #362, opened 2024-02-29
Conditional Providers: Only use providers that are configured properly

julienduchesne:
My use case: Instead of failing because mimir is not configured, or SM, only use the Grafana provider 
A warning is output to mention that mimir and SM won't be used

### DIFF
--- a/cmd/grr/main.go
+++ b/cmd/grr/main.go
@@ -33,29 +33,24 @@ func main() {
 	if err != nil {
 		log.Fatalln(err)
 	}
+	providerInitFuncs := map[string]func() (grizzly.Provider, error){
+		"Grafana":              func() (grizzly.Provider, error) { return grafana.NewProvider(&context.Grafana) },
+		"Mimir":                func() (grizzly.Provider, error) { return mimir.NewProvider(&context.Mimir) },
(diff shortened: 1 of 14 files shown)


K-Phoen (approved):

cmd/grr/main.go
```diff
@@ -33,12 +33,30 @@ func main() {
...
-		[]grizzly.Provider{
-			grafana.NewProvider(&context.Grafana),
-			mimir.NewProvider(&context.Mimir),
-			syntheticmonitoring.NewProvider(&context.SyntheticMonitoring),
-		})
+	providers := []grizzly.Provider{}
+
+	grafanaProvider, err := grafana.NewProvider(&context.Grafana)
+	if err != nil {
+		log.Warnf("Grafana provider is not configured: %v", err)
```
malcolmholmes, line 40: Do we want to warn, or silently ignore? I always feel uncomfortable Grizzly listing the ignored providers. Perhaps we could list the *configured* providers rather than the ignored ones?

malcolmholmes: I often feel uncomfortable seeing the skipped handlers. I told it I don't want to use them, so strictly speaking it isn't "skipping" them. Same deal here?

malcolmholmes (approved):

malcolmholmes (approved):

PR #364, opened 2024-03-01
Add datasource integration tests

malcolmholmes:
- Almost working
- Final test passes

### DIFF
--- a/integration/datasource_test.go
+++ b/integration/datasource_test.go
@@ -12,6 +12,71 @@ import (
 )
 
 func TestDatasources(t *testing.T) {
+	dir := "testdata/datasources"
+	setupContexts(t, dir)
+
+	t.Run("Get datasource - success", func(t *testing.T) {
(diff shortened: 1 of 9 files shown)


integration/datasource_test.go
```diff
@@ -12,6 +12,74 @@ import (
...
+			Commands: []Command{
+				{
+					Command:        "apply datasource-broken.json",
+					ExpectedOutput: "Datasource.datasource-broken added\n",
+				},
+				{
+					Command: "get -o json Datasource.datasource-broken",
+					ExpectedOutputFiles: []string{
+						"datasource-broken-returned.json",
+						"datasource-broken-returned2.json",
```
julienduchesne, line 75: I don't understand. Why is there the same file twice with a different name?

malcolmholmes: because the subdir instance returns different data. But for now I've just excluded the subdir target for this test for this reason.

PR #367, opened 2024-03-05
Add gotestsum options to makefile

malcolmholmes:
Go test does not report failure numbers. When making big changes this can make it hard to
analyse the output.

This PR adds `gotestsum` options to the makefile. It does this instead of replacing the
standard `go test` options to avoid a required dependency.﻿

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,8 +20,15 @@ test-clean:
 test:
 	go test -v ./cmd/... ./pkg/...
 
+testsum:
+	gotestsum ./cmd/... ./pkg/...
+
 integration: run-test-image-locally dev
 	go test -v ./integration/...
+	make stop-test-image-locally
+
+integrationsum: run-test-image-locally dev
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,10 +20,17 @@ test-clean:
 test:
 	go test -v ./cmd/... ./pkg/...
```
julienduchesne, line 21: How about
```
GOTEST ?= "go test"

test:
	$(GOTEST) -v ./cmd/... ./pkg/...
```

Then you can you can run it like this: `GOTEST="gotestsum" make test"

It feels weird because this doesn't do anything different other than replacing go test with an external exec

malcolmholmes: The -v must be inside the GOTEST=.

The downside of this approach is it means I need to remember the syntax. I guess the makefile can have a comment (or a target that sets GOTEST?)

PR #373, opened 2024-03-11
Add a grr self-update command

K-Phoen:
This PR adds a basic `self-update` command, to make it easier to keep grizzly up-to-date.

Later, we could improve our release pipeline to publish checksums of our binary and verify these checksums in the self-update process.


julienduchesne (approved):
Nice 👍

PR #374, opened 2024-03-12
grizzly/actions/setup GitHub action

K-Phoen:
This PR introduces a GitHub action that can be used to setup Grizzly in a job.

I'm thinking that having two github actions could be nice:
* one to setup the binary and nothing else, giving complete freedom to the user
* a second to make "apply" actions easier (it could install grizzly, help configuring it and wrap the `apply` command)

Both these actions would make it easier to build automation on top of grizzly, in GitHub actions.

Open questions:

* where should these actions be published? Having them in this repository is convenient, but makes versioning them much harder


julienduchesne (commented):
I feel like we could use a `go install` composite action and avoid a lot of javascript here

Also, do we have somewhere we want to use this or is this for future use?

K-Phoen:
> I feel like we could use a go install composite action and avoid a lot of javascript here

Running `go install` in a composite action would require us (or the caller) to setup go first, and then we would download Grizzly's dependencies and re-compile the binary every time this action is called.

I wanted to ensure fast runtimes by downloading pre-compiled binaries instead, and taking advantage of GitHub action's "tools cache". That last part is why this action is written in Javascript.

Unfortunately, I also just discovered that this cache isn't persistent across workflow runs :( (see https://github.com/actions/toolkit/issues/58)
That probably means that a composite option that downloads the binary from our releases is a valid option again.

> Also, do we have somewhere we want to use this or is this for future use?

Nothing specific in mind, this PR is just me fooling around :sweat_smile: 
I'm still guessing that a "grizzly/setup" action could be a useful thing to offer, for everyone running grizzly in their github pipelines.

Even though it could be made slightly more useful by providing ways to setup Grizzly's configuration/contexts. Something like:

```yaml
- uses: grafana/grizzly/actions/setup@main
  with:
    version: 'v0.4.0'
    context:
      grafana.url: https://my-instance.grafana.net
      grafana.token: ${{ secrets.GRAFANA_TOKEN }}

- run: grr diff ./resources
```

Admittedly, this action isn't absolutely necessary to use grizzly in a github action. I just feel like we could provide it to remove some of the boilerplate necessary to set it up :shrug: 

WDYT?

BrendanCoughlan5:
+1 please land this, i'd love it :)

julienduchesne (approved):

BrendanCoughlan5:
plz land this xD

mhixon4479:
+1 to landing this :)

K-Phoen:
We released this GitHub action in its own repository: https://github.com/grafana/setup-grizzly

PR #384, opened 2024-03-14
Document 0.4.0 features

malcolmholmes:


Makefile
```diff
@@ -48,3 +48,8 @@ cross: $(GOX)
 # Docker container
 container: static
 	docker build -t grafana/grizzly .
+
+docs:
```
julienduchesne, line 52: `make docs` sounds like it generates docs (as make's default function is to make things). I'd call it `serve-docs`

docs/content/_index.md
```diff
@@ -1,21 +1,48 @@
...
+more.
 
-Grizzly empowers you to provision dashboards using the outputs of your existing as-code workflows.
-Whether you use Terraform or Pulumi, Waypoint or Dagger, Grizzly is the essential last-mile for
-declarative, as-code observability of your existing workflows.
+Grizzly can be a valuable component in a number of workflows around Grafana resources.
 
-***Code, deploy and observe, together.***
+## Edit/Publish
+![Edit workflow](images/workflow-edit.png)
```
julienduchesne, line 16: Why is there a `2. ...` in that image when it's the first one being presented?

docs/content/server.md
```diff
@@ -0,0 +1,28 @@
+---
+date: 2024-03-14
+title: Grizzly Server
+---
+## An HTTP Server for editing and reviewing
+
+Do you have Grafana resources, dashboards, datasources, alerts, etc, on disk,
+and want a way to edit or review them before publishing?
+
+Grizzly Server is here to help.
```
julienduchesne, line 10: Feel free to ignore this if you don't agree 😄 : Personally, I like my docs drier and less telemarkety

malcolmholmes: Will downplay. But that's bigtime what this PR does in general with the docs - they're way OTT at the moment.

docs/content/_index.md
malcolmholmes: Bah. They were originally one image with three stories. I should commit the excalidraw files for these so they can be edited. I'll remove the numbers.

julienduchesne (approved):

PR #385, opened 2024-03-14
Set the UID of existing rules when updating an alert group.

### LINKED ISSUE #381: [Bug]: Updating alert rule groups does not seem to work
### Grizzly Version

0.4.0

### Expected Behavior

`grr apply resource` should work when the resource is an `AlertRuleGroup` that contains rules that already exist on the server.  Even when those rules do not have UIDs inside them, since the UIDs can be determined by matching the names of the existing rules.

### Actual Behavior

`grr apply resource` does not work when the resource is a `AlertRuleGroup` that contains rules that already exist on the server.  The error message is `FATA[0000] fetching alert rule: json: cannot unmarshal array into Go value of type models.ProvisionedAlertRule`

### Steps to Reproduce

_No response_

chrismgrayftsinc:
This uses the heuristic of the name being the same to determine if two rules are the same.  The same heuristic is used within grafana to reject alert rules in the same folder that have the same name as having a conflict, so this seems reasonable to me.

Fixes #381


pkg/grafana/alertgroup-handler.go
```diff
@@ -212,18 +212,52 @@ func (h *AlertRuleGroupHandler) updateAlertRule(rule *models.ProvisionedAlertRul
...
+	}
+	return &group, nil
+}
+
+func updateResourceToHaveUids(existing, resource grizzly.Resource) (*models.AlertRuleGroup, error) {
+	group, err := unMarshalAlertRuleGroup(existing)
+	if err != nil {
+		return nil, err
+	}
+	t := make(map[*string]string)
```
theSuess, line 233: This needs to be a `map[string]string`.

By using pointers, the lookup below will always fail as the same value will have a different pointer

pkg/grafana/alertgroup-handler.go
```diff
@@ -212,18 +212,52 @@ func (h *AlertRuleGroupHandler) updateAlertRule(rule *models.ProvisionedAlertRul
...
 	err = json.Unmarshal(data, &group)
 	if err != nil {
-		return err
+		return nil, err
+	}
+	return &group, nil
+}
+
+func updateResourceToHaveUids(existing, resource grizzly.Resource) (*models.AlertRuleGroup, error) {
+	group, err := unMarshalAlertRuleGroup(existing)
```
theSuess, line 229:
```suggestion
	existingGroup, err := unMarshalAlertRuleGroup(existing)
```

IMHO it's better to have the different groups as separate variables to keep things clear

pkg/grafana/alertgroup-handler.go
```diff
@@ -212,18 +212,52 @@ func (h *AlertRuleGroupHandler) updateAlertRule(rule *models.ProvisionedAlertRul
...
+	if err != nil {
+		return nil, err
+	}
+	t := make(map[*string]string)
+	for _, rule := range group.Rules {
+		t[rule.Title] = rule.UID
+	}
+
+	group, err = unMarshalAlertRuleGroup(resource)
```
theSuess, line 238:
```suggestion
	updatedGroup, err = unMarshalAlertRuleGroup(resource)
```

pkg/grafana/alertgroup-handler.go
```diff
@@ -212,18 +212,52 @@ func (h *AlertRuleGroupHandler) updateAlertRule(rule *models.ProvisionedAlertRul
...
+
+	uidsAdded := false
+	for _, rule := range group.Rules {
+		if uid, ok := t[rule.Title]; ok {
+			rule.UID = uid
+			uidsAdded = true
+		}
 	}
+	if !uidsAdded {
+		return nil, errors.New("No UIDs to add")
```
theSuess, line 251: This approach works for some usecases but fails when updating an alert rule group by adding a new rule. In this case, there is no existing rule to take the UID from, so `uidsAdded` will be false

I'd go for checking if every rule has a UID and fail if one does not

theSuess (changes requested):
Thanks for the contribution!

I've tested some more involved scenarios and found some issues in the code. Once they are fixed, I'm more than happy to merge this

pkg/grafana/alertgroup-handler.go
chrismgrayftsinc: Actually, how would that work?  This function is called from the PUT use-case, which does a GET by UID.  So if the UID doesn't exist, that GET fails.  Should we add a check before that GET also?

I'm not totally sure that this particular check is all that helpful any more -- we could probably just remove it.

pkg/grafana/alertgroup-handler.go
chrismgrayftsinc: Done

pkg/grafana/alertgroup-handler.go
chrismgrayftsinc: Done.  (Though it was working in my testing.  But I am no expert on golang pointers)

pkg/grafana/alertgroup-handler.go
chrismgrayftsinc: Done

pkg/grafana/alertgroup-handler.go
```diff
@@ -212,18 +212,52 @@ func (h *AlertRuleGroupHandler) updateAlertRule(rule *models.ProvisionedAlertRul
 	return err
 }
 
-func (h *AlertRuleGroupHandler) putAlertRuleGroup(resource grizzly.Resource) error {
-	// TODO: Turn spec into a real models.AlertRuleGroup object
+func unMarshalAlertRuleGroup(resource grizzly.Resource) (*models.AlertRuleGroup, error) {
```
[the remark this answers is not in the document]
chrismgrayftsinc: Done

pkg/grafana/alertgroup-handler.go
theSuess: The reason why this fails is that we operate on a `AlertRuleGroup` level - not singular alert rules. This can result in the scenario where:

Remote:
* Rule A
Grizzly:
* Rule B

In this case, the rule applied from Grizzly will not have a UID added and the `No UIDs to add` error is returned.

I think removing the check is fine - what was your original intent when adding this check?

chrismgrayftsinc: I actually don't remember the original intention. This PR started out as a patch on a patch that added alert functionality before this handler existed.

I will remove this check, though I do think it would be worthwhile to check the existence of every rule for a UID before we call the GET.

chrismgrayftsinc: I have now added that second check.

pkg/grafana/alertgroup-handler.go
```diff
@@ -194,13 +194,22 @@ func (h *AlertRuleGroupHandler) updateAlertRule(rule *models.ProvisionedAlertRul
...
+				return h.createAlertRule(rule)
+			}
+			return fmt.Errorf("fetching alert rule: %w", err)
 		}
-		return fmt.Errorf("fetching alert rule: %w", err)
+	} else {
+		stringtrue := "true"
+		params := provisioning.NewPostAlertRuleParams().
+			withBody(rule).
+			withXDisableProvenance(&stringtrue)
```
theSuess, line 210: The functions are public so they need to be uppercase
```suggestion
		params := provisioning.NewPostAlertRuleParams().
			WithBody(rule).
			WithXDisableProvenance(&stringtrue)
```

pkg/grafana/alertgroup-handler.go
```diff
@@ -194,13 +194,22 @@ func (h *AlertRuleGroupHandler) updateAlertRule(rule *models.ProvisionedAlertRul
 		return err
 	}
 
-	_, err = client.Provisioning.GetAlertRule(rule.UID)
-	if err != nil {
-		var gErr *provisioning.GetAlertRuleNotFound
-		if errors.As(err, &gErr) {
-			return h.createAlertRule(rule)
+	if rule.UID != nil {
```
theSuess, line 197: `string` cannot be compared to `nil`
```suggestion
	if rule.UID != "" {
```

theSuess (changes requested):
I found three small issues, but once they are fixed, I'm happy to merge this!

theSuess (approved):

PR #386, opened 2024-03-14
Add live reload functionality

malcolmholmes:
Currently, Grizzly Server does not notice when a file is edited. This PR attempts to add a
livereload feature. When a file is changed on disk (similar to `grr watch` which it will
likely replace), the dashboard will be reloaded.

Grafana itself has "live" functionality. This PR wires itself into that. Support is currently for dashboards only.


julienduchesne:
> Grafana itself has its "live" functionality that could be used to achieve that. However,
> that is currently for dashboards only. This PR instead attempts to add generic live reload
> support by "stealing" code from [Hugo](https://github.com/gohugoio/hugo).

The PR description is out of date, right? We do use the `live` feature

malcolmholmes:
> > Grafana itself has its "live" functionality that could be used to achieve that. However,
> > that is currently for dashboards only. This PR instead attempts to add generic live reload
> > support by "stealing" code from [Hugo](https://github.com/gohugoio/hugo).
> 
> The PR description is out of date, right? We do use the `live` feature

Correct - I just updated the description. Thx

julienduchesne:
Basic auth fix here: https://github.com/grafana/grizzly/pull/432

julienduchesne (commented):
Works great! Just the one comment about auth validation

julienduchesne (approved):

PR #391, opened 2024-03-20
Reinstate preview as snapshot cmd

malcolmholmes:
PR #349 removed the preview feature, as it was an old (and believed unused) feature. After
discussion with colleagues and some users, we propose to re-add it, but with a more
on-target name: `snapshot`, given that previewing something should not have side-effects.﻿

### DIFF
--- a/cmd/grr/main.go
+++ b/cmd/grr/main.go
@@ -78,6 +78,7 @@ func main() {
 		applyCmd(registry),
 		watchCmd(registry),
 		exportCmd(registry),
+		snapshotCmd(registry),
 		providersCmd(registry),
 		configCmd(),
 		serveCmd(registry),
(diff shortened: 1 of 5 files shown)


julienduchesne (approved):

malcolmholmes:
Does this work for either of you? @tekicode @mikaello

mikaello:
> Does this work for either of you? @tekicode @mikaello

Works! And thanks a lot for being so forthcoming!

tekicode:
> Does this work for either of you? @tekicode @mikaello

Thank you, works great!

mikaello:
Would it make sense to also include the documentation removed in https://github.com/grafana/grizzly/pull/384 ? I.e.:

> ### grr snapshot
> When a backend supports snapshot functionality, this renders Jsonnet and
> uploads snapshots to endpoint systems.
> 
> At present, only Grafana dashboards are supported. With Grafana, it produces
> dashboard snapshots. It then prints out links for each snapshot that was uploaded.
> 
> ```sh
> $ grr snapshot my-lib.libsonnet
> ```
> Grafana snapshots by default do not expire. Expiration can be set via the
> `-e, --expires` flag which takes a number of seconds as an argument.

Issue #393, opened 2024-03-22
[Bug]: Nil pointer exception when pulling synthetic checks


dmanc:
### Grizzly Version

0.4.0

### Expected Behavior
```
➜  grizzly git:(main) ✗ ./grr pull -d resources --output json -t SyntheticMonitoringCheck
INFO[0000] Providers:
  Grafana - active
  Mimir - inactive: exec: "cortextool": executable file not found in $PATH
  Synthetic Monitoring - active 
Flag --directory has been deprecated, now it is inferred from the operating system
INFO[0000] Pulling resources to resources               
AlertNotificationPolicy skipped
AlertContactPoint skipped
Pulling 1 resources
SyntheticMonitoringCheck.Test Monitor pulled
```
### Actual Behavior

```
➜  grizzly git:(main) ✗ ./grr pull -d resources --output json -t SyntheticMonitoringCheck
INFO[0000] Providers:
  Grafana - active
  Mimir - inactive: exec: "cortextool": executable file not found in $PATH
  Synthetic Monitoring - active 
Flag --directory has been deprecated, now it is inferred from the operating system
INFO[0000] Pulling resources to resources               
AlertContactPoint skipped
Pulling 1 resources
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x102f69784]

goroutine 1 [running]:
github.com/grafana/grizzly/pkg/grizzly.(*BaseHandler).APIVersion(...)
        /Users/dmanc/github/grizzly/pkg/grizzly/handler.go:26
github.com/grafana/grizzly/pkg/syntheticmonitoring.(*SyntheticMonitoringHandler).getRemoteCheck(0x140004f3d70, {0x1400046ebc0, 0x1d})
        /Users/dmanc/github/grizzly/pkg/syntheticmonitoring/synthetic-monitoring-handler.go:225 +0x454
github.com/grafana/grizzly/pkg/syntheticmonitoring.(*SyntheticMonitoringHandler).GetByUID(0x0?, {0x1400046ebc0?, 0x18?})
        /Users/dmanc/github/grizzly/pkg/syntheticmonitoring/synthetic-monitoring-handler.go:115 +0x20
[...]

PR #394, opened 2024-03-22
Fix nil pointer when pulling SyntheticMonitoringCheck

dmanc:
Fixes https://github.com/grafana/grizzly/issues/393

### DIFF
--- a/pkg/syntheticmonitoring/synthetic-monitoring-handler.go
+++ b/pkg/syntheticmonitoring/synthetic-monitoring-handler.go
@@ -211,7 +211,6 @@ func (h *SyntheticMonitoringHandler) getRemoteCheck(uid string) (*grizzly.Resour
 			for _, probeID := range check.Probes {
 				probeNames = append(probeNames, probes.ByID[probeID].Name)
 			}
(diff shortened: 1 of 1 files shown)


pkg/syntheticmonitoring/synthetic-monitoring-handler.go
```diff
@@ -211,7 +211,7 @@ func (h *SyntheticMonitoringHandler) getRemoteCheck(uid string) (*grizzly.Resour
 			for _, probeID := range check.Probes {
 				probeNames = append(probeNames, probes.ByID[probeID].Name)
 			}
-			handler := SyntheticMonitoringHandler{}
+			handler := NewSyntheticMonitoringHandler(h.Provider)
```
julienduchesne, line 214: I feel like we should use the current struct (`h`) rather than create a new handler

dmanc: I see, made the change.

BrendanCoughlan5:
+1 looks good to me :)  :) :)

julienduchesne (approved):

PR #395, opened 2024-03-24
Update configuration.md

moxious:


julienduchesne (commented):
Not so sure about this. I think this a bug rather than something we should document. The directory should be created automatically IMO

PR #396, opened 2024-03-26
Fix "save" via server and track provenance

### LINKED ISSUE #378: [Feature Request]: Can `save` button on Grafana GUI, save to yaml file directly?
### Feature Request

I am a newbie to this repository, I tried to search similar issues but could not find, or could not make it work if there is already this functionality.

It would be really nice addition if the "`grr serve` for dashboards" supports saving to the yaml file directly. 
Since `yaml` files are great for checking git differences, we are leaning towards to use it in our QA process. But editing within yaml file is cumbersome for non developers in development process. 

So as a user and code reviewer, I would expect builtin save button in the dev server web interface to change files within served folder. 

Optionally we can override dashboard `tags` values just before saving to file since they are good facility to distinguish grr managed dashboards on Grafana.

malcolmholmes:
Currently, when Grizzly parses resources, it does not track where those resources came from. Without this,
writing back to resources becomes impossible.

This PR adds source support, and while at it, fixes "save" so that dashboards saved via the proxy will work correctly.

Closes #378.

### DIFF
--- a/cmd/grr/workflow.go
+++ b/cmd/grr/workflow.go
@@ -45,7 +45,9 @@ func listCmd(registry grizzly.Registry) *cli.Command {
 	}
 	var opts Opts
 	var isRemote bool
+	var isVerbose bool
 	cmd.Flags().BoolVarP(&isRemote, "remote", "r", false, "list remote resources")
+	cmd.Flags().BoolVarP(&isVerbose, "verbose", "v", false, "show more information about resources")
 
 	cmd.Run = func(cmd *cli.Command, args []string) error {
 		currentContext, err := config.CurrentContext()
@@ -80,7 +82,7 @@ func listCmd(registry grizzly.Registry) *cli.Command {
 			return err
 		}
 
-		return grizzly.List(registry, resources)
+		return grizzly.List(registry, resources, isVerbose)
 	}
 	return initialiseCmd(cmd, &opts)
 }
--- a/pkg/grafana/dashboard-handler.go
+++ b/pkg/grafana/dashboard-handler.go
@@ -403,18 +403,18 @@ func (h *DashboardHandler) DashboardJSONPostHandler(p grizzly.Server) http.Handl
 
 		out, _, _, err := grizzly.Format(p.Registry, p.ResourcePath, &resource, p.OutputFormat, p.OnlySpec)
 		if err != nil {
-			grizzly.SendError(w, "Error formatting content", err, 400)
+			grizzly.SendError(w, "Error formatting content", err, 500)
 			return
 		}
 
 		resources, err := p.Parser.Parse()
 		if err != nil {
-			grizzly.SendError(w, "Error parsing existing resources", err, 400)
+			grizzly.SendError(w, "Error parsing existing resources", err, 500)
 			return
 		}
 		existing, found := resources.Find(grizzly.NewResourceRef("Dashboard", uid))
 		if !found {
-			grizzly.SendError(w, fmt.Sprintf("Dashboard with UID %s not found", uid), fmt.Errorf("dashboard with UID %s not found", uid), 404)
+			grizzly.SendError(w, fmt.Sprintf("Dashboard with UID %s not found", uid), fmt.Errorf("dashboard with UID %s not found", uid), 500)
 			return
 		}
 		if !existing.Source.Rewritable {
@@ -424,7 +424,7 @@ func (h *DashboardHandler) DashboardJSONPostHandler(p grizzly.Server) http.Handl
 
 		err = os.WriteFile(existing.Source.Path, out, 0644)
 		if err != nil {
-			grizzly.SendError(w, fmt.Sprintf("Error writing file: %s", err), err, 400)
+			grizzly.SendError(w, fmt.Sprintf("Error writing file: %s", err), err, 500)
 			return
 		}
 
--- a/pkg/grafana/dashboard-handler.go
+++ b/pkg/grafana/dashboard-handler.go
@@ -353,8 +353,9 @@ func (h *DashboardHandler) DashboardJSONGetHandler(p grizzly.Server) http.Handle
 			grizzly.SendError(w, fmt.Sprintf("Dashboard with UID %s not found", uid), fmt.Errorf("dashboard with UID %s not found", uid), 404)
 			return
 		}
-		resource.SetSpecValue("version", 1)
-
+		if resource.GetSpecValue("version") == nil {
+			resource.SetSpecValue("version", 1)
+		}
 		meta := map[string]interface{}{
 			"type":      "db",
 			"isStarred": false,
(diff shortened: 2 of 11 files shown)


cmd/grr/workflow.go
```diff
@@ -45,7 +45,9 @@ func listCmd(registry grizzly.Registry) *cli.Command {
 	}
 	var opts Opts
 	var isRemote bool
+	var isVerbose bool
 	cmd.Flags().BoolVarP(&isRemote, "remote", "r", false, "list remote resources")
+	cmd.Flags().BoolVarP(&isVerbose, "verbose", "v", false, "show more information about resources")
```
K-Phoen, line 50: Having both a `log-level` and a `verbose` flag makes be :cry: and :confused:

pkg/grizzly/json.go
```diff
@@ -34,7 +34,7 @@ func (parser *JSONParser) Parse(file string, options ParserOptions) (Resources,
 		return nil, err
 	}
 
-	resources, err := parseAny(parser.registry, m, options.DefaultResourceKind, options.DefaultFolderUID)
+	resources, err := parseAny(parser.registry, m, options.DefaultResourceKind, options.DefaultFolderUID, "json", file, true)
```
K-Phoen, line 37: The signature of this function feels a bit overwhelming. From here, it's also not obvious what `true` means here.

pkg/grizzly/resources.go
```diff
@@ -30,6 +30,13 @@ func (ref ResourceRef) String() string {
 // Resource represents a single Resource destined for a single endpoint
 type Resource struct {
 	Body map[string]interface{}
+
+	Provenance struct {
```
K-Phoen, line 34: Why using an anonymous type here? Having a `Provenance` type would also simplify the signature of `SetProvenance()`

pkg/grizzly/resources.go
```diff
@@ -30,6 +30,13 @@ func (ref ResourceRef) String() string {
...
+		Format     string
+		Location   string
+		Path       string
+		Rewritable bool
```
K-Phoen, line 38: This seems to be always dependent on the format, is storing this even necessary? Couldn't a "writer" decide whether to actually write depending on the format?

pkg/grizzly/resources.go
```diff
@@ -30,6 +30,13 @@ func (ref ResourceRef) String() string {
 // Resource represents a single Resource destined for a single endpoint
 type Resource struct {
 	Body map[string]interface{}
+
+	Provenance struct {
+		Format     string
+		Location   string
```
K-Phoen, line 36: What's the difference between `Location` and `Path`? :confused:

pkg/grizzly/resources.go
malcolmholmes: At first, a json resource with multiple objects wouldn't be writable. But you're saying decide at writing time not at parse time.

pkg/grizzly/json.go
malcolmholmes: yah, time for a struct.

malcolmholmes: Actually, with a `Source` struct, this gets much simpler.

cmd/grr/workflow.go
malcolmholmes: They are totally different though. One shows different logs, this tells Grizzly to report more back to the user about the resources. Do you have any suggestions for a different flag name?

pkg/grizzly/resources.go
malcolmholmes: Heh, one is `FilePath` and the other is `LocationWithinFile`. Better names? Suggestions?

pkg/grizzly/utils.go
```diff
@@ -13,3 +16,8 @@ func Pluraliser(count int, name string) string {
 
 	return fmt.Sprintf("%d %ss", count, name)
 }
+
+func SendError(w http.ResponseWriter, msg string, err error, code int) {
+	http.Error(w, msg, 400)
```
K-Phoen, line 21:
```suggestion
	http.Error(w, msg, code)
```

cmd/grr/workflow.go
K-Phoen: Ideally, I'd keep only `log-level` (even though it usually is named `verbosity`, with different levels) and use a `format` flag where needed.

Verbosity representing the level of debugging info being printed out (ie: what is the CLI doing), and format controlling the format of the information being printed.

A bit like what `kubectl` does

K-Phoen: In this PR, I would skip the verbose flag entirely and always print the source.

Cleaning up the way logs and actual info is handled in grizzly sounds a bit too big to be done here :)

pkg/grafana/dashboard-handler.go
```diff
@@ -372,36 +375,56 @@ func (h *DashboardHandler) DashboardJSONGetHandler(p grizzly.Server) http.Handle
...
 		out, _, _, err := grizzly.Format(p.Registry, p.ResourcePath, &resource, p.OutputFormat, p.OnlySpec)
 		if err != nil {
-			http.Error(w, "Error formatting content", 400)
+			grizzly.SendError(w, "Error formatting content", err, 400)
+			return
+		}
+
+		resources, err := p.Parser.Parse()
+		if err != nil {
+			grizzly.SendError(w, "Error parsing existing resources", err, 400)
```
K-Phoen, line 412:
```suggestion
			grizzly.SendError(w, "Error parsing existing resources", err, 500)
```

Same here. Though ideally we wouldn't need to parse everything here (this should be changed... eventually)

pkg/grafana/dashboard-handler.go
```diff
@@ -372,36 +375,56 @@ func (h *DashboardHandler) DashboardJSONGetHandler(p grizzly.Server) http.Handle
...
+			return
+		}
+		existing, found := resources.Find(grizzly.NewResourceRef("Dashboard", uid))
+		if !found {
+			grizzly.SendError(w, fmt.Sprintf("Dashboard with UID %s not found", uid), fmt.Errorf("dashboard with UID %s not found", uid), 404)
```
K-Phoen, line 417:
```suggestion
			grizzly.SendError(w, fmt.Sprintf("Dashboard with UID %s not found", uid), fmt.Errorf("dashboard with UID %s not found", uid), 500)
```

pkg/grafana/dashboard-handler.go
```diff
@@ -372,36 +375,56 @@ func (h *DashboardHandler) DashboardJSONGetHandler(p grizzly.Server) http.Handle
...
+		if !existing.Source.Rewritable {
+			grizzly.SendError(w, "The source for this dashboard is not rewritable", fmt.Errorf("the source for this dashboard is not rewritable"), 400)
 			return
 		}
 
-		err = os.WriteFile(p.ResourcePath, out, 0644)
+		err = os.WriteFile(existing.Source.Path, out, 0644)
 		if err != nil {
-			http.Error(w, fmt.Sprintf("Error writing file: %s", err), 400)
+			grizzly.SendError(w, fmt.Sprintf("Error writing file: %s", err), err, 400)
```
K-Phoen, line 427:
```suggestion
			grizzly.SendError(w, fmt.Sprintf("Error writing file: %s", err), err, 500)
```

K-Phoen (approved):

PR #398, opened 2024-03-26
Create the grizzly config directory if needed

K-Phoen:
Rather than documenting the potential need to create Grizzly's config directory: let's have Grizzly do the work.

Could replace #395

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -219,9 +219,18 @@ func Write() error {
 		return err
 	}
 
+	// Ensure that our configuration directory exists: viper only takes care of
+	// creating the file.
+	configDir := configdir.LocalConfig("grizzly")
+	if _, err := os.Stat(configDir); os.IsNotExist(err) {
+		if err := os.Mkdir(configDir, 0700); err != nil {
+			return err
+		}
+	}
+
 	// Viper failed because no configuration file exists in the "config path".
 	// We explicitly tell it where to write its config: at the most global location.
-	globalConfigPath := filepath.Join(configdir.LocalConfig("grizzly"), "settings.yaml")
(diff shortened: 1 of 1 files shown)


julienduchesne (approved):
👍 Thanks!

pkg/config/config.go
```diff
@@ -219,9 +219,18 @@ func Write() error {
 		return err
 	}
 
+	// Ensure that our configuration directory exists: viper only takes care of
+	// creating the file.
+	configDir := configdir.LocalConfig("grizzly")
+	if _, err := os.Stat(configDir); os.IsNotExist(err) {
+		if err := os.Mkdir(configDir, 0700); err != nil {
```
malcolmholmes, line 226: `os.Mkdir` or `os.MkdirAll`?

malcolmholmes (approved):

pkg/config/config.go
K-Phoen: On a linux machine, `configDir` is `~/.config/grizzly`

`os.Mkdir()` lets us create the `grizzly` folder if it doesn't exist while `os.MkdirAll()` would potentially create everything in that path (which practically should never be needed anyway). I decided against using `MkdirAll()` for two reasons: we shouldn't create things we don't own, it would create directories with 0700 permissions (which is probably not what is usually set on `~/.config`)

marinnedea: by default ~/.config directory doesn't exists on most Linux distributions by default.

Without that, you'll run into:
```
$ grr config create-context grafana-onprem
FATAL mkdir /home/<user>/.config/grizzly: no such file or directory
```

I think `os.MkdirAll()` is better for this. Or, change the requirement for the config to use a multiple directories path.. so instead of `~/.config/grizzly` maybe `~/grizzly` is enough.

PR #399, opened 2024-03-28
Check for nil folder UID

### LINKED ISSUE #263: "grr get DashboardFolder." leads to unmarshall error
Trying to export a list of folders from the server, and it cannot unmarshall. Output below was redacted.

```
$ grr get DashboardFolder.
INFO[0000] Getting DashboardFolder.                     
FATA[0000] Error retrieving dashboard folder : Failed to parse Grafana response: json: cannot unmarshal array into Go value of type grafana.Folder.

Response:
[{"id":n,"uid":"x","title":"x"},...,{"id":n,"uid":"x","title":"x"}]
```

malcolmholmes:
Closes #263﻿

### DIFF
--- a/pkg/grafana/folder-handler.go
+++ b/pkg/grafana/folder-handler.go
@@ -138,6 +138,9 @@ func (h *FolderHandler) Update(existing, resource grizzly.Resource) error {
 
 // getRemoteFolder retrieves a folder object from Grafana
 func (h *FolderHandler) getRemoteFolder(uid string) (*grizzly.Resource, error) {
+	if uid == "" {
+		return nil, fmt.Errorf("No folder UID provided")
+	}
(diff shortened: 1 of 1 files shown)


pkg/grafana/folder-handler.go
```diff
@@ -151,6 +151,9 @@ func (h *FolderHandler) getRemoteFolder(uid string) (*grizzly.Resource, error) {
 		if err != nil {
 			return nil, err
 		}
+		if uid == "" {
```
julienduchesne, line 154: Should be the very first condition of the function IMO

malcolmholmes: fair. Moved.

julienduchesne (approved):

PR #400, opened 2024-04-01
Adds provider's name and small refactor

spinillos:
* It adds `Name()` function to each provider.
* It fixes silentError check using `errors.As`.
* Registry creation was moved into a function.

### DIFF
--- a/cmd/grr/main.go
+++ b/cmd/grr/main.go
@@ -22,9 +22,8 @@ type silentError struct {
 }
 
 func (err silentError) Is(target error) bool {
-	_, ok := target.(silentError)
-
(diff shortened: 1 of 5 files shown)


cmd/grr/main.go
```diff
@@ -94,3 +70,29 @@ func main() {
 		}
 	}
 }
+
+func createRegistry() grizzly.Registry {
+	context, err := config.CurrentContext()
```
malcolmholmes, line 75: We want to pass context in rather than pull it from a global. Can you handle this?

malcolmholmes (approved):

PR #402, opened 2024-04-02
Move providers information into config

spinillos:
To archive that we needed to change the way to create the providers. The constructor doesn't check the missing configuration and it was moved into a `validate` function. This function is called internally when we do want to do queries to the resources.

Now configuration has two new options:
```bash
grr config providers handlers
grr config providers status
```

_Maybe `validate` needs a better name since is exposed to make `status` to work._

* `handlers` is the old `grr providers` command but showing all providers information instead of the active ones (the ones with the full configuration set).
* `status` shows the information that it showed all the time when you execute any command showing if the provider is active or not.

### DIFF
--- a/test-docker-compose/docker-compose.yml
+++ b/test-docker-compose/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3'
+version: '3.4'
 services:
   grafana:
     image: &image grafana/grafana:10.2.0
--- a/cmd/grr/config.go
+++ b/cmd/grr/config.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"fmt"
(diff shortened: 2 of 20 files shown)


malcolmholmes:
I wonder it "kinds" is a better name than "handlers", with handlers being an internal name. What people will care about is the kinds that Grizzly supports (and the providers they must configure).

test-docker-compose/docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3'
+version: '3.4'
```
spinillos, line 1: Its added because `start_period` works from `3.4`

PR #403, opened 2024-04-03
Adds an interface for cortex tool

spinillos:
It adds an interface for CortexTool to avoid to be able to mock the behaviour in tests. Currently its overriding the whole function to make it works.

### DIFF
--- a/pkg/mimir/cortex_tool.go
+++ b/pkg/mimir/cortex_tool.go
@@ -1,0 +1,39 @@
+package mimir
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/grafana/grizzly/pkg/config"
+)
+
+type CortexTool interface {
+	ExecuteCortexTool(args ...string) ([]byte, error)
+}
+
+type Cortex struct {
+	config *config.MimirConfig
+}
+
+func NewCortexTool(config *config.MimirConfig) *Cortex {
+	return &Cortex{config: config}
+}
+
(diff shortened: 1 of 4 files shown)


pkg/mimir/cortex_tool.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/grafana/grizzly/pkg/config"
+)
+
+type CortexTool interface {
+	ExecuteCortexTool(args ...string) ([]byte, error)
```
julienduchesne, line 12: I think it should be named in a way where we could switch over to mimirtool by default. 

Maybe it should be abstracted more so that it's a "prometheus client" and instead of `ExecuteCortexTool` we have `GetRules` that gives us a `map[string][]PrometheusRuleGroup{}` and a `LoadRules` function, etc, and so then we can use whatever tool we want in the backend

julienduchesne (commented):
I like the idea. I have a comment about making the Interface show less of the actual implementation

pkg/mimir/cortex_tool.go
spinillos: I did it as a wrapper to be able to "mock" `exec.LookPath` behaviour and use it in tests easily.

But its a good idea because I'm working in a small client that its doing exactly that you mention 😂.  Mimir exposes an API and it could be a good idea to be able to use any of the options.

julienduchesne (approved):
Good that you're working on it! Approving, because as it is, this PR is better than what's currently in `main`.

PR #406, opened 2024-04-09
Create mimir client

spinillos:
This PR adds the possibility to use `mimirtool` or `http` client for Mimir.

To archive that, cortex interface was modified to show it as a real client adding `ListRules` and `LoadRules` functions.
Additionally, the configuration allows you to add mimir/cortex paths by `config set` and a `mimir.client` to select the desired tool to use with Mimir.

After a few tests, `auth-key` isn't mandatory for local environments and `tenant-id` is mandatory for all of the tools. Also `tenant-id` was changed to string because for `http` and `mimirtool` isn't a integer.


julienduchesne (commented):
Just one question: What is the benefit of using cortex tool or mimirtool if we can use http directly. Seems like the external binaries are inferior and we could remove the options if they are

spinillos:
I wanted to leave `cortextool` to avoid to break current implementation if any was using it. It should be deprecated in favor of `mimirtool`. 

`http` doesn't allow to send multiple rule groups at once, when `mimirtool` allows you to do it. I don't know if `http` is more limited than `mimirtool` and its why I added both.

spinillos:
I talked with Oleg from Mimir and he told me that its completely fine to use only the http client. So if you agree to use only http client and remove current `cortextool` implementation, its ok for me.

pkg/config/config.go
```diff
@@ -39,9 +39,9 @@ func override(v *viper.Viper) {
 		"synthetic-monitoring.logs-id":    "GRAFANA_SM_LOGS_ID",
 		"synthetic-monitoring.metrics-id": "GRAFANA_SM_METRICS_ID",
 
-		"mimir.address":   "CORTEX_ADDRESS",
-		"mimir.tenant-id": "CORTEX_TENANT_ID",
```
julienduchesne, line 43: Can we keep retrocompatibility (support two envvars at the same time)? We may have people using these envvars with Mimir or Grafana Cloud. This is a breaking change

pkg/mimir/client/http_client.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot read response body: %s", err)
+	}
+
+	return b, nil
+}
+
+func createHttpClient() (*http.Client, error) {
+	timeout := 10 * time.Second
+	if timeoutStr := os.Getenv("GRIZZLY_HTTP_TIMEOUT"); timeoutStr != "" {
```
julienduchesne, line 121: Can you add a TODO here to make this a config item (or do it in another preparation PR)? Getting an envvar in multiple locations in the codebase is weird when we have central config management. The duration could be parsed centrally as well

pkg/mimir/errors.go
```diff
@@ -0,0 +1,11 @@
+package mimir
+
+import "fmt"
+
+type ErrNoBinarySet struct {
```
julienduchesne, line 5: This isn't used, it seems

docs/content/configuration.md
```diff
@@ -149,13 +150,21 @@ docs](https://grafana.com/docs/grafana/latest/http_api/auth/) for more info.
 ## Grafana Cloud Prometheus
 To interact with Grafana Cloud Prometheus, you must have these environment variables set:
 
+<<<<<<< HEAD
+| Name               | Description | Required |
+|--------------------| --- |----------|
+| `MIMIR_ADDRESS`    | URL for Grafana Cloud Prometheus instance | true     |
+| `MIMIR_TENANT_ID` | Tenant ID for your Grafana Cloud Prometheus account | true     |
+| `MIMIR_API_KEY`   | Authentication token/api key | false    |
+=======
```
malcolmholmes, line 159: oops :-)

pkg/config/config.go
```diff
@@ -39,18 +39,37 @@ func override(v *viper.Viper) {
...
-		"mimir.api-key":   "CORTEX_API_KEY",
+		"mimir.address":   "MIMIR_ADDRESS",
+		"mimir.tenant-id": "MIMIR_TENANT_ID",
+		"mimir.api-key":   "MIMIR_API_KEY",
 	}
+
+	// To keep retro compatibility
+	mimirBindings := map[string]string{
```
malcolmholmes, line 48: could we just call these `legacyBindings` then they'll work for all providers?

malcolmholmes (approved):

docs/content/configuration.md
spinillos: 😅 Fixed

PR #411, opened 2024-04-11
Keep resources in memory in the proxy

K-Phoen:
Instead of parsing the entire resource collection every time we need to access a single resource, let's parse it once when the proxy starts and give the handlers access to that collection.

### DIFF
--- a/pkg/grizzly/server.go
+++ b/pkg/grizzly/server.go
@@ -163,7 +163,7 @@ func (p *Server) Start() error {
 			return fmt.Errorf("no resources found to proxy")
 		}
 
-		if !stat.IsDir() && p.Resources.Len() != 0 {
+		if !stat.IsDir() && p.Resources.Len() == 1 {
 			resource := p.Resources.First()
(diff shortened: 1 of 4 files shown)


pkg/grizzly/server.go
```diff
@@ -141,53 +147,71 @@ func (p *Server) Start() error {
...
 
-		if stat.IsDir() {
-			url = fmt.Sprintf("http://localhost:%d", p.Port)
-		} else {
-			resources, err := p.Parser.Parse()
+		if !stat.IsDir() && p.Resources.Len() == 0 {
+			return fmt.Errorf("no resources found to proxy")
+		}
+
+		if !stat.IsDir() && p.Resources.Len() != 0 {
```
malcolmholmes, line 166:
```suggestion
		if !stat.IsDir() && p.Resources.Len() == 1 {
```
Jsonnet can render multiple resources from a single file (or multiple resources can be in the same file). So we only want to open a specific resource in the browser if there's only one resource.

malcolmholmes (approved):
Approving with one minor tweak.

PR #412, opened 2024-04-11
Support loading dashboards into the proxy via a query parameter

K-Phoen:
With this PR, the proxy/server can be started without an initial resources catalog. Resources can then be dynamically parsed with a _magic_ query parameter when opening a dashboard.

Ex: opening `http://localhost:4242/d/JvOqE4gRk/slug?orgId=1&refresh=30s&grizzly_from_file=/home/kevin/sandbox/dashboard.json` will parse and display the dashboard defined in the file `/home/kevin/sandbox/dashboard.json`


julienduchesne (approved):
That's cool 👍

PR #418, opened 2024-04-18
Add option to 'get' directly to a file

malcolmholmes:
Currently, `grr get` prints the output. This PR adds a `-O` option to write the output to a 
named file.

This can work well, for example, for prepping work with the server/proxy:

```
grr get -O mydash.json -o json Dashboard.mydash
grr proxy mydash.json
```

### DIFF
--- a/cmd/grr/workflow.go
+++ b/cmd/grr/workflow.go
@@ -24,6 +24,8 @@ func getCmd(registry grizzly.Registry) *cli.Command {
 		Args:  cli.ArgsExact(1),
 	}
 	var opts Opts
+	var output string
+	cmd.Flags().StringVarP(&output, "output-file", "O", "", "write resource to specific file")
 
 	cmd.Run = func(cmd *cli.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


julienduchesne (commented):
Hmm, what about piping to a file instead? I mean, do we need to add this feature?

cmd/grr/workflow.go
```diff
@@ -24,14 +24,16 @@ func getCmd(registry grizzly.Registry) *cli.Command {
 		Args:  cli.ArgsExact(1),
 	}
 	var opts Opts
+	var output string
+	cmd.Flags().StringVarP(&output, "output-file", "O", "", "write resource to specific file")
```
spinillos, line 28: Maybe is better to use `-out`? Having `-O` and `-o` could be confusing.

K-Phoen:
+1 with @julienduchesne: redirecting the output into a file seems good enough to me :shrug: 

Closing this, but feel free to reopen the PR if I'm missing something :)

Issue #415, opened 2024-04-16
Setup Mimir integration testing environment


spinillos:
### Feature Request

Setup Mimir's integration environment and add some tests to ensure that it works.

PR #419, opened 2024-04-18
Adds Mimir integration tests

spinillos:
Contributes to https://github.com/grafana/grizzly/issues/415

It adds Mimir configuration to docker-compose and an two integration tests to send and retrieve the rules.

The test adds one ugly second of delay between tests because Mimir takes some time to sync the rules. Setting `poll_interval` to `1s` in rules configuration, speeds up the sync, but we still need this small delay to avoid 500 error responses.

✅ ~**Note**: It could be nice have this https://github.com/grafana/grizzly/pull/406 merged first to be able to add more use cases and use http client.~


test-docker-compose/nginx.conf
```diff
@@ -10,6 +10,10 @@ http {
     upstream grafana {
         server grafana_subpath:3002;
     }
+    
+    upstream backend {
```
malcolmholmes, line 14:
```suggestion
    upstream mimir {
```

integration/mimir/testdata/test-rules-1.yml
```diff
@@ -0,0 +1,16 @@
+apiVersion: grizzly.grafana.com/v1alpha1
+kind: RuleGroup1
```
malcolmholmes, line 2: doesn't `RuleGroup1` confuse Grizzly? Shouldn't it be `PrometheusRuleGroup`? Same for `RuleGroup2` and `RuleGroup3` below.

malcolmholmes (approved):
These are great tests!

Issue #388, opened 2024-03-14
[Bug]: Grizzly doesn't see any SyntheticMonitoringChecks


asreich:
### Grizzly Version

0.4.0

### Expected Behavior

We have configured 86 Synthetic Monitoring Checks via the UI and Grizzly should be able to find them and pull down a JSON representation of them as described [here](https://grafana.github.io/grizzly/synthetic-monitoring/)

### Actual Behavior

Grizzly finds 0 SyntheticMonitoringCheck resources

### Steps to Reproduce

```
export GRAFANA_URL=https://mycompany.grafana.net
export GRAFANA_TOKEN=<SNIP>

# Generate Access policy with metrics/logs read/write and stack:read
grr config set synthetic-monitoring.token <SNIP> # API key (must have MetricsPublisher permissions)
grr config set synthetic-monitoring.stack-id <SNIP>
grr config set synthetic-monitoring.metrics-id <SNIP>
grr config set synthetic-monitoring.logs-id <SNIP>

➜  ~/Downloads/grr-0.4.0 pull -d resources --output json -t SyntheticMonitoringCheck
Flag --directory has been deprecated, now it is inferred from the operating system
INFO[0000] Pulling resources to resources               
LibraryElement skipped
AlertRuleGroup skipped
AlertContactPoint skipped
PrometheusRuleGroup skipped
Datasource skipped
Dashboard skipped
AlertNotificationPolicy skipped
No resources found
Pulling 0 resources
DashboardFolder skipped
```

julienduchesne:
Probably a duplicate of https://github.com/grafana/grizzly/issues/216

asreich:
Yeah, after reviewing [this](https://grafana.com/docs/grafana-cloud/monitor-public-endpoints/set-up/set-up-private-probes/#probe-api-server-url) page and comparing the hardcoded address to the us-east endpoint, that seems accurate. Feel free to close this in favor of the prior if you like.

spinillos:
`grr config set synthetic-monitoring.url` was added to set your own URL.

PR #420, opened 2024-04-19
Add region to SM backend URL

spinillos:
Related: https://github.com/grafana/grizzly/issues/388

It allows to set the region for Synthetic Monitoring.

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -58,7 +58,6 @@ grr config set synthetic-monitoring.logs-id # Logs instance ID
 grr config set synthetic-monitoring.region # Region of your backend address.
 ```
 Your stack ID is the number at the end of the url when you view your Grafana instance details, ie. `grafana.com/orgs/myorg/stacks/123456` would be `123456`. Your metrics and logs ID's are the `User` when you view your Prometheus or Loki instance details in Grafana Cloud.
-You can check the regions under Synthetic Monitoring configuration. For example, if your backend URL is `synthetic-monitoring-api-gb-south.grafana.net`, your region should be `gb-south`. Default value is `synthetic-monitoring-api.grafana.net`.
 
 ## Configuring Targets
 Grizzly supports a number of resource types (`grr providers` will list those supported). Often, however, we do not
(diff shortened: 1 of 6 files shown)


docs/content/configuration.md
```diff
@@ -135,37 +137,39 @@ In some circumstances (e.g. when used within automated pipelines) it makes sense
...
+|-------------------------|-----------------------------------------------------------------------|----------|
+| `GRAFANA_SM_TOKEN`      | Authentication token/api key (must have MetricsPublisher permissions) | true     |
+| `GRAFANA_SM_STACK_ID`   | Grafana instance/stack ID                                             | true     |
+| `GRAFANA_SM_LOGS_ID`    | Logs instance ID                                                      | true     |
+| `GRAFANA_SM_METRICS_ID` | Metrics instance ID                                                   | true     |
+| `GRAFANA_SM_REGION`     | Region of your instance                                               | false    |
 
 Your stack ID is the number at the end of the url when you view your Grafana instance details, ie. `grafana.com/orgs/myorg/stacks/123456` would be `123456`. Your metrics and logs ID's are the `User` when you view your Prometheus or Loki instance details in Grafana Cloud.
+You can check the regions under Synthetic Monitoring configuration. For example, if your backend URL is `synthetic-monitoring-api-gb-south.grafana.net`, your region should be `gb-south`. Default value is `synthetic-monitoring-api.grafana.net`.
```
malcolmholmes, line 172: Is `synthetic-monitoring-api.grafana.net` the default? Or `us-central-1` or such?

malcolmholmes (approved):

pkg/syntheticmonitoring/synthetic-monitoring-handler.go
```diff
@@ -26,6 +26,7 @@ import (
  */
 
 const smBaseURL = "https://synthetic-monitoring-api.grafana.net"
+const smRegionURL = "https://synthetic-monitoring-api-%s.grafana.net"
```
julienduchesne, line 29: I would add the same exceptions as here: https://github.com/grafana/terraform-provider-grafana/blob/main/internal/resources/cloud/resource_synthetic_monitoring_installation.go#L16-L22

Otherwise, if someone creates a stack in `eu`, and tries to use Grizzly, it will fail. All new APIs are being provisioned with the API URL you have here

docs/content/configuration.md
julienduchesne: It's the value for `us`. My other comment will make that clearer

julienduchesne (commented):
Suggestion on

pkg/syntheticmonitoring/synthetic-monitoring-handler.go
spinillos: I'm seeing that the list [here](https://grafana.com/docs/grafana-cloud/monitor-public-endpoints/set-up/set-up-private-probes/) is larger than the one set in TF provider.

malcolmholmes: That's probes, not API endpoints, right? A probe is the remote node that pings your host. They connect to an API for configuration. There will be far fewer API endpoints than probes.

pkg/syntheticmonitoring/provider.go
```diff
@@ -11,6 +11,14 @@ import (
 	smapi "github.com/grafana/synthetic-monitoring-api-go-client"
 )
 
+var smAPIURLsExceptions = map[string]string{
```
julienduchesne, line 14: This isn't used

julienduchesne (approved):
LGTM, just two small comments

PR #428, opened 2024-05-01
Added mTLS support to MimirClient

### LINKED ISSUE #426: [Feature Request]: Support client ceritifacte authentication for Mimir
### Feature Request
Add support for client ceritifacte authenticaiton to Mimir via cli
`grr config set mimir.tls.client-certificate-path /path/to/cert`
`grr config set mimir.tls.client-key-path /path/to/key`

We want to utilize this tool in our environment, we provide Mimir behind a gateway secured with client certificates.

Itaykal:
Added optional mTLS config to Mimir client.
Added an option to append an additional ca bundle file to mimir client.

I think it should wait for #415 as the changes could not be tested at the moment.

Fixes #426

### DIFF
--- a/pkg/config/model.go
+++ b/pkg/config/model.go
@@ -9,15 +9,16 @@ type GrafanaConfig struct {
 }
 
 type MimirConfig struct {
-	Address  string 		 `yaml:"address" mapstructure:"address"`
-	TenantID string 	     `yaml:"tenant-id" mapstructure:"tenant-id"`
-	APIKey   string          `yaml:"api-key" mapstructure:"api-key"`
-	Auth 	 MimirAuthConfig `yaml:"auth" mapstructure:"auth"`
+	Address  string         `yaml:"address" mapstructure:"address"`
+	TenantID string         `yaml:"tenant-id" mapstructure:"tenant-id"`
+	APIKey   string         `yaml:"api-key" mapstructure:"api-key"`
+	TLS      MimirTLSConfig `yaml:"tls" mapstructure:"auth"`
 }
 
-type MimirAuthConfig struct {
-	ClientCertPath string `yaml:"client-cert-path" mapstructure:"client-cert-path"`
-	ClientKeyPath  string `yaml:"client-key-path" mapstructure:"client-key-path"`
+type MimirTLSConfig struct {
+	ClientCertPath string `yaml:"client-cert-path,omitempty" mapstructure:"client-cert-path"`
+	ClientKeyPath  string `yaml:"client-key-path,omitempty" mapstructure:"client-key-path"`
+	CAPath 		   string `yaml:"ca-path" mapstruceture:"ca-path"`
 }
 
 type SyntheticMonitoringConfig struct {
--- a/pkg/config/model.go
+++ b/pkg/config/model.go
@@ -12,7 +12,7 @@ type MimirConfig struct {
 	Address  string         `yaml:"address" mapstructure:"address"`
 	TenantID string         `yaml:"tenant-id" mapstructure:"tenant-id"`
 	APIKey   string         `yaml:"api-key" mapstructure:"api-key"`
-	TLS      MimirTLSConfig `yaml:"tls" mapstructure:"auth"`
+	TLS      MimirTLSConfig `yaml:"tls" mapstructure:"tls"`
 }
 
 type MimirTLSConfig struct {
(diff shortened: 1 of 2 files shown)


pkg/config/model.go
```diff
@@ -9,9 +9,16 @@ type GrafanaConfig struct {
 }
 
 type MimirConfig struct {
-	Address  string `yaml:"address" mapstructure:"address"`
-	TenantID string `yaml:"tenant-id" mapstructure:"tenant-id"`
-	APIKey   string `yaml:"api-key" mapstructure:"api-key"`
+	Address  string         `yaml:"address" mapstructure:"address"`
+	TenantID string         `yaml:"tenant-id" mapstructure:"tenant-id"`
+	APIKey   string         `yaml:"api-key" mapstructure:"api-key"`
+	TLS      MimirTLSConfig `yaml:"tls" mapstructure:"auth"`
```
malcolmholmes, line 15: Is there a reason for yaml being `tls` and mapstructure being `auth`?

pkg/config/model.go
```diff
@@ -9,9 +9,16 @@ type GrafanaConfig struct {
...
+}
+
+type MimirTLSConfig struct {
+	ClientCertPath string `yaml:"client-cert-path,omitempty" mapstructure:"client-cert-path"`
+	ClientKeyPath  string `yaml:"client-key-path,omitempty" mapstructure:"client-key-path"`
+	CAPath 		   string `yaml:"ca-path" mapstruceture:"ca-path"`
```
malcolmholmes, line 21:
```suggestion
	CAPath 		   string `yaml:"ca-path" mapstructure:"ca-path"`
```

malcolmholmes (commented):
This is an amazing contribution, thanks! I've left a couple of comments, but will also await comment from @spinillos as this touches her code.

pkg/config/model.go
Itaykal: I forgot to change it :) fixing

pkg/mimir/client/http_client.go
```diff
@@ -126,5 +128,42 @@ func createHTTPClient() (*http.Client, error) {
...
+		}
+
+		caCertPEM, err := os.ReadFile(c.config.TLS.CAPath)
+		if err != nil {
+			return nil, err
+		} 
+		
+		ok := certPool.AppendCertsFromPEM(caCertPEM)
+		if !ok {
+			return nil, err
```
spinillos, line 151: Error here is nil, you should return a custom error.

pkg/mimir/client/http_client.go
```diff
@@ -126,5 +128,42 @@ func createHTTPClient() (*http.Client, error) {
...
+		}
+		
+		tlsConfig.RootCAs = certPool
+	}
+
+	if c.config.TLS.ClientCertPath != "" || c.config.TLS.ClientKeyPath != "" {
```
spinillos, line 157: Does it work if we only set one of them? In that case, should it be `&&`?

pkg/config/model.go
```diff
@@ -9,9 +9,16 @@ type GrafanaConfig struct {
...
+	Address  string         `yaml:"address" mapstructure:"address"`
+	TenantID string         `yaml:"tenant-id" mapstructure:"tenant-id"`
+	APIKey   string         `yaml:"api-key" mapstructure:"api-key"`
+	TLS      MimirTLSConfig `yaml:"tls" mapstructure:"tls"`
+}
+
+type MimirTLSConfig struct {
+	ClientCertPath string `yaml:"client-cert-path,omitempty" mapstructure:"client-cert-path"`
+	ClientKeyPath  string `yaml:"client-key-path,omitempty" mapstructure:"client-key-path"`
+	CAPath 		   string `yaml:"ca-path" mapstructure:"ca-path"`
```
spinillos, line 21: It isn't formatted.

spinillos (commented):
Thanks for your contribution 😃!! I added a few comments.

pkg/mimir/client/http_client.go
Itaykal: `LoadX509KeyPair` should return an error if the file path doesn't exist and if the contents are not a valid certificate, is this enough or should I return an error when only one of them is configured?

spinillos: I think that it's enough. Thanks for the clarification.

spinillos (approved):

PR #431, opened 2024-05-01
`grr serve`: Require one argument

julienduchesne:
From my tests, it doesn't seem to do anything if you don't give an arg It should be restricted then


malcolmholmes:
This is needed for embedding into an IDE (such as @K-Phoen is exploring) when there is no initial set of resources to review.

julienduchesne:
> This is needed for embedding into an IDE (such as @K-Phoen is exploring) when there is no initial set of resources to review.

hmm, can we print a warning then?

malcolmholmes:
Hmmm, or use a switch (`-n` or `--no-resources`) to tell it that you don't have anything to parse. Without the switch, throw an error.

PR #434, opened 2024-05-02
Add remote resource viewing support

malcolmholmes:
The Grizzly server allows users to see local resources, i.e. on the local workstation.

This PR adds support for viewing remote resources (as seen by `grr list -r`). On top of that, it offers the ability to see those resources as YAML/JSON.

There's plenty of UI/UX opportunities here. This PR (currently) just shows what's possible.

Another possibility is offering the ability to "download" or rather "save" resources to disk. They would then show up in the local view. My only concern here is that that isn't expected behaviour of a web page so might confuse people.

### DIFF
--- a/pkg/grizzly/embed/assets/style.css
+++ b/pkg/grizzly/embed/assets/style.css
@@ -64,3 +64,8 @@ main {
     margin-right: auto;
     max-width: 61rem;
 }
+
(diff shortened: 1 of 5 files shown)


malcolmholmes:
image
image
image
image

PR #435, opened 2024-05-07
Validate auth on server start

malcolmholmes:
At present, the Grizzly server will start, even if the Grafana creds are
invalid. Problems only show up when attempting to view a web page.

This PR adds a simple authentication check when establishing the proxy,
to confirm authentication, and give prompt feedback if all is not
extablished correctly.

### DIFF
--- a/pkg/grafana/provider.go
+++ b/pkg/grafana/provider.go
@@ -3,12 +3,14 @@ package grafana
 import (
 	"crypto/tls"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"net/http/httputil"
 	"net/url"
 	"path/filepath"
 
 	gclient "github.com/grafana/grafana-openapi-client-go/client"
+	"github.com/grafana/grafana-openapi-client-go/client/dashboards"
 	"github.com/grafana/grizzly/pkg/config"
 	"github.com/grafana/grizzly/pkg/grizzly"
 )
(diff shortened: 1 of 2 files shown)


pkg/grafana/provider.go
```diff
@@ -105,6 +105,15 @@ func (p *Provider) GetHandlers() []grizzly.Handler {
 }
 
 func (p *Provider) SetupProxy() (*httputil.ReverseProxy, error) {
+	client, err := p.Client()
+	if err != nil {
+		return nil, err
+	}
+	_, err = client.Dashboards.GetHomeDashboard()
```
spinillos, line 112: I understand that it's checking any kind of error here. I mean, at least that we want to fail anyway its ok but the error returned should be different.

If we want to specify that the error is because the authentication is invalid, we should cast the error with [this](https://github.com/grafana/grafana-openapi-client-go/blob/main/client/dashboards/get_home_dashboard_responses.go#L33-L38).

So in that case should be something like:
```go
if err != nil {
  if errors.Is(err, &dashboards.GetHomeDashboardUnauthorized{}) {
    return nil, fmt.Errorf("error checking authentication: %v", err)
  }
  return nil, fmt.Errorf("error setting the proxy: %v", err)
}
```

malcolmholmes: applied

spinillos (approved):

PR #436, opened 2024-05-08
Allow offline providers

### LINKED ISSUE #424: [Bug]: Mandatory provider for offline commands
### Grizzly Version

0.4.1

### Expected Behavior

`podman run --rm -v $PWD/:/src:Z -w /src grafana/grizzly:0.4.1 show foo.jsonnet` should show rendered results. It should not require a connection to Grafana because it's purely offline, which enables local development and CI usage without passing information for the later environment. 
```
time="2024-04-25T14:03:00Z" level=info msg="Showing 28 resources"
DashboardFolder.foo:
...
```

Correct results are returned when downgrading: `podman run --rm -v $PWD/:/src:Z -w /src grafana/grizzly:0.4.0 show foo.jsonnet`.

### Actual Behavior

Instead it prints:
> Providers: Grafana - inactive (grafana URL is not set), Mimir - inactive (exec: "cortextool": executable file not found in $PATH), Synthetic Monitoring - inactive (stack id is not set)
time="2024-04-25T14:16:45Z" level=fatal msg="parse error in 'foo.jsonnet': couldn't find a handler for DashboardFolder: handler not found"

### Steps to Reproduce

The container provides the correct environment. The file contents:
```json
{
  "apiVersion": "grizzly.grafana.com/v1alpha1",
  "kind": "DashboardFolder",
  "metadata": {
    "name": "foo"
  },
  "spec": {
    "title": "foo"
  }
}
```
The breaking change can be fixed by passing any contents to environment variable GRAFANA_URL, e.g. with `-e GRAFANA_URL=foo`. I assume #362 to be the origin.

malcolmholmes:
#362 added details of which providers were configured. It also prevented unconfigured
providers from being added to the list of available providers.

This has the unfortunate consequence of preventing "offline" commands from working,
e.g. `grr show`, `grr list` or `grr export`.

This PR resolves this by still allowing an unconfigured provider to be used for such
situations. Now, an unconfigured provider will just error out when it attempts to
connect to its backend.

Closes #424.

### DIFF
--- a/cmd/grr/main.go
+++ b/cmd/grr/main.go
@@ -81,17 +81,14 @@ func main() {
 }
 
 func createRegistry(context *config.Context) grizzly.Registry {
-	providerInitFuncs := []func() grizzly.Provider{
-		func() grizzly.Provider { return grafana.NewProvider(&context.Grafana) },
-		func() grizzly.Provider { return mimir.NewProvider(&context.Mimir) },
-		func() grizzly.Provider { return syntheticmonitoring.NewProvider(&context.SyntheticMonitoring) },
+	providers := []grizzly.Provider{
+		grafana.NewProvider(&context.Grafana),
+		mimir.NewProvider(&context.Mimir),
+		syntheticmonitoring.NewProvider(&context.SyntheticMonitoring),
 	}
 
-	var providers []grizzly.Provider
-
 	var providerList []string
-	for _, initFunc := range providerInitFuncs {
-		provider := initFunc()
+	for _, provider := range providers {
 		err := provider.Validate()
 		if err != nil {
 			providerList = append(providerList, fmt.Sprintf("%s - inactive (%s)", provider.Name(), err.Error()))
--- a/cmd/grr/main.go
+++ b/cmd/grr/main.go
@@ -81,22 +81,23 @@ func main() {
 }
 
 func createRegistry(context *config.Context) grizzly.Registry {
-	providerInitFuncs := []func() (grizzly.Provider, error){
-		func() (grizzly.Provider, error) { return grafana.NewProvider(&context.Grafana) },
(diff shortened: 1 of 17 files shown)


julienduchesne:
I think this needs a few tests. It should be easy to write an integration test that checks that the offline commands work without any configured providers and it should prevent regressions (no one caught it in my PR, and no one probably will, without tests, a year from now)

malcolmholmes:
@julienduchesne integration test added!

cmd/grr/main.go
```diff
@@ -81,22 +81,23 @@ func main() {
...
 func createRegistry(context *config.Context) grizzly.Registry {
-	providerInitFuncs := []func() (grizzly.Provider, error){
-		func() (grizzly.Provider, error) { return grafana.NewProvider(&context.Grafana) },
-		func() (grizzly.Provider, error) { return mimir.NewProvider(&context.Mimir) },
-		func() (grizzly.Provider, error) { return syntheticmonitoring.NewProvider(&context.SyntheticMonitoring) },
+	providerInitFuncs := []func() grizzly.Provider{
+		func() grizzly.Provider { return grafana.NewProvider(&context.Grafana) },
+		func() grizzly.Provider { return mimir.NewProvider(&context.Mimir) },
+		func() grizzly.Provider { return syntheticmonitoring.NewProvider(&context.SyntheticMonitoring) },
 	}
```
julienduchesne, line 88:
```suggestion
	providers := []grizzly.Provider{
		grafana.NewProvider(&context.Grafana),
		mimir.NewProvider(&context.Mimir),
		syntheticmonitoring.NewProvider(&context.SyntheticMonitoring),
	}
```
Doesn't need to be a func if we're validating in a separate func

julienduchesne (approved):
Just one nit. LGTM!

PR #437, opened 2024-05-09
Add server toolbar

malcolmholmes:
A dashboard, when viewed via the Grizzly Server, doesn't look any different to any other dashboard. This can lead to confusion as to what people are actually seeing.

This PR adds a consistent header across the two pages served by the server - the "root" page that lists all dashboards, and the dashboard view. On the dashboard view, it first identifies that this is running via the Grizzly Server, then it identifies the current context (as this is often the cause of dashboards not working - you're using a context that lacks the necessary datasources), and an `i` icon pointing to the Grizzly docs about the server (which need improving).


pkg/grizzly/server.go
```diff
@@ -40,7 +41,7 @@ var upgrader = websocket.Upgrader{
 	CheckOrigin:     func(r *http.Request) bool { return true },
 }
 
-func NewGrizzlyServer(registry Registry, parser Parser, parserOpts ParserOptions, resourcePath string, port int, openBrowser, watch, onlySpec bool, outputFormat string) (*Server, error) {
+func NewGrizzlyServer(registry Registry, resourcePath string, port int) (*Server, error) {
```
spinillos, line 44: Looking at the changes, this could follow the "functional options pattern",  WDYT?

malcolmholmes: It would, except, I'd need to move the server to its own package, otherwise we'd get `grizzly.withContext()` rather than `server.withContext()`.

spinillos (approved):

PR #447, opened 2024-05-17
Add option to configure Mimir alertmanager

### LINKED ISSUE #387: [Feature Request]: External Alertmanager Contact Point and Notification Policy Support
### Feature Request

Thanks to the newly added `AlertRuleGroup,AlertContactPoint,AlertNotificationPolicy` targets we can pull these new resources but not from configurations that use an External Alertmanager to handle Grafana Cloud Alert Rules. The AlertRuleGroups can be synced easily enough but all of the Contact points and Notification Policies I have setup in the external Alertmanager cannot be.

My use case is migrating from on-prem Grafana OSS to Grafana Cloud and Grizzly is making it super easy. I can manually create these Contact Points and Notification Policies to make things work but just wanted to highlight a gap that would be hugely helpful for other people's migration prospects.

Thanks for your work on this project!

Itaykal:
fixes #387 

This PR fixes some comments, namings and object specific errors (which shouldn't be this specific)


malcolmholmes:
Thanks for this contribution! We can look at it next week. In the meantime, there's a few lint type things noted in the PR. Can you review them?

Itaykal:
> Thanks for this contribution! We can look at it next week. In the meantime, there's a few lint type things noted in the PR. Can you review them?

Will do when possible, please note that the changes are not final, some tests are missing and some features like `list -r` are behaving in a weird manner.

Another important note that I think will need some further discussion here is that there's no name and/or uid for such object and I truly don't know how to handle it.

malcolmholmes:
> Another important note that I think will need some further discussion here is that there's no name and/or uid for such object and I truly don't know how to handle it.

How did @theSuess handle it for the equivalent Grafana resource? I think the pattern is already there.

Itaykal:
> How did @theSuess handle it for the equivalent Grafana resource? I think the pattern is already there.

Good point, seems like the implementation is based on having a static name/uid for said object, I could try to align my implementation with theirs.

Itaykal:
Another issue I'm currently having is that `pkg/mimir/rules_test.go` needs changing as I changed the mimir client, is this test even relevant? I see that the rules are being tested as part of the integration tests under `integration/mimir/rules_test.go`

skinlayers:
Any chance of resurrecting this PR? We can't use Grizzly for our most common use cases without it.
Thank you!

Itaykal:
> Any chance of resurrecting this PR? We can't use Grizzly for our most common use cases without it. Thank you!

I don't know if i'll have much time to do that soon :/
Just started a new position and this PR stopped being relevant to me, tough I would like to keep working on it in the future if I get to it.

I'd recommend to checkout from my branch and add the missing pieces

PR #459, opened 2024-06-24
Use appropriate authentication header for basicauth in Grafana provider

aabramov-ionos:
Hi team!

Currently Grizzly tries to send `Authentication: Bearer <base64(username:password)>` instead of `Authentication: Basic <base64(username:password)>` even when basicauth is explicitly required by specifying `config.User` and `config.Token`.
This didn't work for me so I've dug into the sources and made this change which fixed my issues. I glanced through Grafana documentation and didn't find any requirement to send the encoded basicauth credentials in bearer header, so I deduced that this is indeed a bug.


julienduchesne (approved):

julienduchesne:
Grafana uses the http.Request BasicAuth() function: https://pkg.go.dev/net/http#Request.BasicAuth. This function takes in `Basic <base64>`, so it is indeed a bug

PR #464, opened 2024-07-12
Support watching jsonnet via server

malcolmholmes:
Currently, the `grr serve` command will only watch its `resourcePath` for ﻿changes. This works fine for a directory full of resources, such as JSON or YAML files. But for Jsonnet, it doesn't.

This PR adds a simple change to allow a user to provide an additional argument, that is, a "watchDir" path, so this command:

```
grr serve -w examples/grr.jsonnet examples
```

will allow a user to edit `examples/test-dashboard.json`, which is included in `grr.jsonnet`, and see that change show immediately in Grafana.

This PR also updates the docs for the Grizzly server, including this change in them.

### DIFF
--- a/cmd/grr/workflow.go
+++ b/cmd/grr/workflow.go
@@ -360,7 +360,7 @@ func serveCmd(registry grizzly.Registry) *cli.Command {
 	cmd := &cli.Command{
 		Use:   "serve <resources>",
 		Short: "Run Grizzly server",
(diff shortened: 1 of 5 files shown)


Duologic (approved):
The additional path is a good quick fix to support jsonnet initially. If it can support multiple directories/files, we could do something like this:
```
grr serve -w examples/grr.jsonnet $(tk tool imports examples/grr.jsonnet)
```

PR #465, opened 2024-07-16
Synthetic Monitoring: Configure with access token

julienduchesne:
Rather than finding stack information and installing SM, a user can get the token directly from the SM config page

The Terraform provider is configured the same way (as well as the config export plugin)


malcolmholmes (approved):
This is great, thanks! Feel free to make a release too.

julienduchesne:
> This is great, thanks! Feel free to make a release too.

I've imported the changeset in the exporter plugin

PR #469, opened 2024-07-30
Fixed push of PrometheusRuleGroup

vladshub:
When doing `grr pull . -t 'PrometheusRuleGroup/*'` and then `grr push . -t 'PrometheusRuleGroup/*'` the format that the API expects in the push is different from what was created in the `pull`.

Fixed this in the push flow, this way if someone uses files that ware already created he would be able to push them.

I also added more information on the errors received from the API calls.


K-Phoen (approved):
LGTM, thanks :muscle: 

Not sure why the CI hasn't run, but I ran the tests locally and everything was green.

PR #473, opened 2024-08-14
Add favicon to Grizzly Server

malcolmholmes:
Shows up in browser tab, which will help to identify the tab when there are many.


pkg/grizzly/server.go
```diff
@@ -314,6 +315,16 @@ func (s *Server) RootHandler(w http.ResponseWriter, _ *http.Request) {
...
 }
 
+func (s *Server) faviconHandlerFunc() http.HandlerFunc {
+	content, _ := embedFS.ReadFile("embed/assets/grizzly.ico")
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "image/x-icon")
+		w.WriteHeader(http.StatusOK)
+		w.Write(content)
+	}
```
Duologic, line 325: Are these lint errors resolvable?

Duologic (approved):

pkg/grizzly/server.go
malcolmholmes: yep, on it.

PR #474, opened 2024-08-14
Server: only parse acceptable files

malcolmholmes:
The server, by default, triggers on every file change. When editing Jsonnet files with `vim`, the server can end up attempting to execute the Jsonnet at a time when vim has not yet moved the file into place. This ends up with weird, "file not found" errors, when the file being reported as not found is the very file you've just been editing.

This is because it triggers on vim's `.swp` file - at that point, the edited file does not actually exist.

This is easily resolved: the parser now has an `Accept()` method, and any file can be checked to see whether Grizzly wants to trigger on it.

Also adds a UID to an example dashboard, which is needed for the server to work correctly.

### DIFF
--- a/pkg/grizzly/server.go
+++ b/pkg/grizzly/server.go
@@ -102,13 +102,14 @@ var mustProxyPOST = []string{
 	"/api/ds/query",
 }
 var blockJSONget = map[string]string{
-	"/api/ma/events":    "[]",
-	"/api/live/publish": "[]",
-	"/api/live/list":    "[]",
-	"/api/user/orgs":    "[]",
-	"/api/annotations":  "[]",
-	"/api/search":       "[]",
-	"/api/usage/*":      "[]",
+	"/api/ma/events":       "[]",
+	"/api/live/publish":    "[]",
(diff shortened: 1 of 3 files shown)


pkg/grizzly/server.go
```diff
@@ -102,13 +102,14 @@ var mustProxyPOST = []string{
...
-	"/api/search":       "[]",
-	"/api/usage/*":      "[]",
+	"/api/ma/events":       "[]",
+	"/api/live/publish":    "[]",
+	"/api/live/list":       "[]",
+	"/api/user/orgs":       "[]",
+	"/api/annotations":     "[]",
+	"/api/search":          "[]",
+	"/api/usage/*":         "[]",
+	"/api/frontend/assets": "{}",
```
Duologic, line 112: I assume this is unrelated and just came up as an endpoint that the proxy needs to capture?

Duologic (approved):

pkg/grizzly/server.go
malcolmholmes: yes

PR #477, opened 2024-08-23
Accept arrays of objects

malcolmholmes:
At present, you can either provide a single object (e.g. a dashboard) to Grizzly, or you provide a tree of objects, containing enveloped resources.

This PR allows users to provide an array of resources, with the type of those resourcs detected by Grizzly. Thus, they can be enveloped, or they can be simple dashboards.

This is a simple, entry-level way to get multiple resources into Grizzly.


Duologic (commented):
Perhaps a (few) testcase(s) would be useful here.

malcolmholmes:
> Perhaps a (few) testcase(s) would be useful here.

tests added.

Duologic (approved):

PR #482, opened 2024-09-01
Fix incorrect argument parsing for serve

malcolmholmes:


cmd/grr/workflow.go
```diff
@@ -379,13 +379,12 @@ func serveCmd(registry grizzly.Registry) *cli.Command {
...
-			}
-			if len(args) > 1 {
-				watchPaths = args[1:]
-			}
+		} else if len(args) == 1 {
+			resourcesPath = args[0]
+			watchPaths = []string{resourcesPath}
+		} else if len(args) > 1 {
+			resourcesPath = args[0]
+			watchPaths = args[1:]
```
Duologic, line 387: This can probably be simplified by assigning a default value:
```go
// grr serve <resourcesPath> [<watchPaths>...]

resourcesPath := args[0]
watchPaths := args

if len(args) > 1 {
	watchPaths = args[1:]
}

if opts.WatchScript != "" {
	resourcesPath = ""
}
```

spinillos (approved):

Issue #492, opened 2024-09-12
[Bug]: 'grr serve -S' without watch does not work


mikaello:
### Grizzly Version

v0.4.7

### Expected Behavior

It should work to run `grr serve -S '<some command>' .` without having to activate watch mode.

### Actual Behavior

No dashboards are loaded in Grizzly UI. If I add `-w` it works (even without touching any other files).

### Steps to Reproduce

1. Export GRAFANA_URL and GRAFANA_TOKEN 
2. `git clone https://github.com/grafana/grafana-foundation-sdk`
3. `cd grafana-foundation-sdk/examples/typescript/red-method`
4. `npm install`
5. `grr serve -S 'npm run -s dev' .`

No dashboards are available:

![Image](https://github.com/user-attachments/assets/85fea4f5-3d91-47c6-b898-0c19906bdf4d)

If I enable watch mode (`-w`) in the command, the dashboard appears:

`grr serve -w -S 'npm run -s dev' .`:

![Image](https://github.com/user-attachments/assets/48a77e7a-553b-4622-9add-b75e05cdb0b2)

malcolmholmes:
Hmm, this is actually what has been implemented. But you're right, `-S` actually requires `-w` - without `-w`, Grizzly will be started with no resources, which is what you are seeing.

mikaello:
So this is intended behavior?

It’s not a big deal for me; I just expected it to work without enabling watch mode. If you don’t plan to change this behavior, feel free to close this issue as "won’t fix."

Duologic:
Looks like intended behavior and should be documented as such.

kimjoonghokim:
I know this is currently assigned to @K-Phoen but since he is OOO, I can pick this up if you think that is alright? @Duologic

PR #495, opened 2024-09-17
Make the watch and watch script options behave more intuitively

K-Phoen:
Related to #492

This PR makes `grr serve` behave in a more intuitive manner when the `-S` flag is used.
With this change, `-S` (watch script) implies `-w` (watch). Furthermore, the given resources path is no longer ignored when a watch script is given.

### DIFF
--- a/pkg/grizzly/server.go
+++ b/pkg/grizzly/server.go
@@ -181,6 +181,11 @@ func (s *Server) Start() error {
 	r.Get("/grizzly/{kind}/{name}", s.IframeHandler)
 	r.Get("/api/live/ws", livereload.LiveReloadHandlerFunc(upgrader))
 
+	_, err = s.ParseResources(s.ResourcePath)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/grizzly/server.go
```diff
@@ -181,18 +181,21 @@ func (s *Server) Start() error {
 	r.Get("/grizzly/{kind}/{name}", s.IframeHandler)
 	r.Get("/api/live/ws", livereload.LiveReloadHandlerFunc(upgrader))
 
+	_, err = s.ParseResources(s.ResourcePath)
+	if err != nil {
+		fmt.Print(err)
+	}
```
malcolmholmes, line 187: Here, you're parsing resources for all invocations. But, with a watchScript, surely the path(s) provided are just to things to watch. We don't want to be parsing `my-dashboard.ts` for example? That's why it doesn't parse when `watchScript` is set.

PR #501, opened 2024-10-10
Config diagnostics

K-Phoen:
While helping a first time grizzly user, we ran into some difficulties trying to identify some configuration issues.

The two changes in this PR would have helped us:

* the first commit introduces a new `grr config check` command, exposing a quick view of what provider is active or not and whether they're online or not (see screenshot below)
* the second one fixes a bug where `grr config get` fails to take environment variables into configuration when printing the current config.

Screenshot from 2024-10-10 17-28-17

### DIFF
--- a/cmd/grr/config.go
+++ b/cmd/grr/config.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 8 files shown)


malcolmholmes (approved):
nice

PR #514, opened 2024-10-30
Make livereloading of watched dashboards more reliable

K-Phoen:
Instead of using Grafana's websocket to trigger a dashboard refresh, let's use a more crude approach and refresh the entire page.

The server-side implementation of the livereload process is taken from https://github.com/gohugoio/hugo/tree/62567d38205a61134a6822d37a534520772419f1/livereload

It has been adjusted to ensure that only dashboards that changed will be reloaded.


pkg/grizzly/server.go
```diff
@@ -289,7 +289,7 @@ func (s *Server) updateWatchedResource(name string) error {
 		}
 		resources, err = s.ParseBytes(b)
 	} else {
-		resources, err = s.ParseResources(s.ResourcePath)
+		resources, err = s.ParseResources(name)
```
K-Phoen, line 292: This is a drive-by fix: let's parse the resource that changed instead of re-parsing all of them.

internal/livereload/connection.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2015 The Hugo Authors. All rights reserved.
```
Duologic, line 1: Can we document the diff-patch we do here? This might make it easier to compare with upstream in the future.

Duologic (approved):
:+1:

internal/livereload/connection.go
K-Phoen: I left a few comments describing where I took the code from and what I changed :)

PR #517, opened 2024-11-04
Fix `grr get` of an AlertRuleGroup with UID containing dots

### LINKED ISSUE #516: [Bug]: AlertRuleGroup get command fails with UID-s containing dot .
### Grizzly Version

 v0.6.0

### Expected Behavior

Get the alert rule group spec

### Actual Behavior

Error : FATAL couldn't find a handler for `AlertRuleGroup.<some-uid>`: handler not found

### Steps to Reproduce

- retrieve alert rule groupes using `grr list -r -t AlertRuleGroup`
- UIDs are formatted like this : `<some-uid>.<some-text>`
- try to get one of these alert rule groups : `grr get AlertRuleGroup.<some-uid>.<some-text>`

K-Phoen:
Fixes #516 

Fixes the logic used by `grr get` to split the Handler name from the rest of the UID.

### DIFF
--- a/integration/alertrules_test.go
+++ b/integration/alertrules_test.go
@@ -1,0 +1,37 @@
+package integration_test
+
+import (
+	"testing"
+)
+
+func TestAlertRuleGroups(t *testing.T) {
+	dir := "testdata/alert-rules"
+	setupContexts(t, dir)
+
+	t.Run("Apply rule group", func(t *testing.T) {
+		runTest(t, GrizzlyTest{
+			TestDir:       dir,
+			RunOnContexts: allContexts,
+			Commands: []Command{
+				{
+					Command:      "apply sample-rule-group.json",
+					ExpectedCode: 0,
+				},
+			},
+		})
+	})
+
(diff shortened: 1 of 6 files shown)


Pokom (approved):
Source code changes looks reasonable enough for me 👍🏻

K-Phoen:
I want to do some more tests tomorrow, from [this bit](https://github.com/grafana/grizzly/pull/517/files#diff-1c2bc400869dd54ad3f708aebc293bb446249797cd4d64de029df22eaf5376fcL28-L39) of code it looks like handler names can be composed of two parts (which surprises me). I feel like a bit more checks are required, in case I break something I'm just not aware of (that also wouldn't be covered by tests :sweat_smile:)

malcolmholmes:
"this bit" above is needed for Synthetic Monitoring, IIRC, as a check didn't have a unique identifier - so we needed to use its name and type to make a unique identifier.

K-Phoen:
Thanks for the tip :)

I tested with SM checks and it seems to work fine.

PR #521, opened 2024-11-12
Escape slashes from resource names when generating filenames

K-Phoen:
For some resources, grizzly uses the resource name/title as UID. This value is then directly used to generate a path on the filesystem under which it will be written.

If this value contains slashes, grizzly will write end up writing the resource at an unexpected location.

To avoid that, this PR replaces slashes by `-` when generating filenames from resources.


malcolmholmes:
The slashes were the seperator before we switched to dots. It should be dots everywhere, but slashes supported for back-compat. Personally, I'd suggest that the switch was so long ago that removing all mention of slashes in the code would be quite reasonable (if back-incompat).

spinillos (approved):

PR #525, opened 2024-11-14
Support Datasources in `grr serve`

K-Phoen:
This PR adds support for datasources in `grr serve`.

Here's how it looks:

<details>
<summary>
Grizzly server home
</summary>

Screenshot 2024-11-14 at 15-42-48 Grizzly

</details>

<details>
<summary>
Datasource page
</summary>

Screenshot 2024-11-14 at 15-43-05 Grizzly

</details>

Note: datasources can NOT be persisted via the grizzly server (yet, this may be implemented later).

### DIFF
--- a/pkg/grafana/dashboard-handler.go
+++ b/pkg/grafana/dashboard-handler.go
@@ -259,22 +259,22 @@ func (h *DashboardHandler) Detect(data map[string]any) bool {
 func (h *DashboardHandler) GetProxyEndpoints(s grizzly.Server) []grizzly.HTTPEndpoint {
 	return []grizzly.HTTPEndpoint{
 		{
-			Method:  "GET",
+			Method:  http.MethodGet,
 			URL:     "/d/{uid}/{slug}",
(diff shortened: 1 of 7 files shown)


malcolmholmes:
This is really cool! So, if my Grafana instance has access to the datasource backends, but the datasources are not configured in Grafana, Grizzly serve can provide them. That's really neat!

K-Phoen:
Not sure what you mean :thinking: 

Are you thinking about using locally-defined datasources when previewing dashboards?
AFAIK, this PR doesn't enable anything like that. It "just" does preview of datasources via `grr serve`, anything else would be an unintended side-effect :sweat_smile:

malcolmholmes:
Hah, okay, so I can preview whether the datasource looks as I would expect via the UI. That's fair. But those unintended side-effects could also be really cool? Maybe?

K-Phoen:
I'm not sure how that would work. Queries are handled by Grafana's backend, making it hard for Grizzly to inject/patch datasources for querying.

We'll have to settle for being able to preview datasources configuration with `grr serve`, and having to properly create datasources in the remote Grafana instance if they're meant to be used in dashboards/alerts/...

spinillos (approved):

PR #554, opened 2025-01-10
Embed docs theme and update it for recent hugo versions

K-Phoen:
Since [hugo got updated](https://github.com/grafana/grizzly/commit/7aa9fc1bc5a77a8cb49b8e71c80e24cf13f61b1c#diff-eb28fbd2c2a71f3e5f1fb977bb8dab63cbe6fa76afe4417001671f648beb76abL20) in our github actions, docs can't be built.

They can't be built because the theme Grizzly is using relies on deprecated variables. That theme being unmaintained, I had to vendor it into this repo to address the build issues :|


Pokom (approved):
@K-Phoen any thoughts on forking the theme ourselves? I don't know how much we want to get into the business of maintaining themes, but we could use this theme for other projects potentially.

PR #560, opened 2025-01-15
Specify steps to create service account for Grafana auth

nikimanoledaki:
Adding more info to the docs on how to authenticate to a Grafana instance using Grizzly.

### DIFF
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -25,7 +25,7 @@ context first, then explain how to use multiple contexts thereafter.
 Settings can be configured via Grizzly itself, using `grr config set`. They are stored
 in an OS specific location.
 
-## Grafana Itself
+## Authenticate with a Grafana instance
 
 Grizzly interacts with Grafana via its REST API. For this, you will need to
 establish authentication credentials.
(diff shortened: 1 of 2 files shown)


jjo (approved):

docs/content/configuration.md
```diff
@@ -25,30 +25,63 @@ context first, then explain how to use multiple contexts thereafter.
...
+
+Optionally, set a username if using [basic authentication](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/#basic-authentication).
+
+```sh
+grr config set grafana.user admin # (Optional) Username if using basic auth
+```
+
+## Authenticate with hosted Prometheus
+
+To interact with [hosted Prometheus / Mimir](./prometheus.md) resources, use these settings:
```
nikimanoledaki, line 73: I replaced `Grafana Cloud Prometheus` with `hosted Prometheus` so that it's clearer that this is specifically for Prometheus/Mimir. That way folks don't wonder whether they need to do this if they want to use Grafana Cloud ☺️

spinillos (approved):
Thank you

PR #564, opened 2025-01-16
Fix the proxy to work as intended even when the targeted resource is …

### LINKED ISSUE #556: [Bug]: Path in root_url in grafana is used two times - Grafana not start
### Grizzly Version

0.6.1

### Expected Behavior

Load grafana correctly when use path in root_url.
Edited:
It should use for proxy only `<url>:<port>`, so dashboard should be accessible in `localhost:8080/<path>/d/<dashboard_id>`, because of settings of links.

### Actual Behavior

I have root url in grafana set to `<url>:<port>/<path>` and when I use Grizzly and set grafana url to `<url>:<port>/<path>`, then dashboard is not loaded and I get error message `If you're seeing this Grafana has failed to load its application files`. 

When I looked at dev tools, I can see, that grizzly want to load resources from `localhost:8080/<path>/<resource>` with error 404. When I try to load source from `localhost:8080/<resource>` it is loaded successfully. 

So it looks like Grizzly use `<path>` part two times, first time when creating "proxy" and second time when it try to load resources.

### Steps to Reproduce

```
grr config create-context grafana
grr config set grafana.url <url>:<port>/<path>
grr config set grafana.token <token>

grr serve -w <dashboard>
```

K-Phoen:
…hosted under a sub-path

Fixes #556

### DIFF
--- a/pkg/grafana/provider.go
+++ b/pkg/grafana/provider.go
@@ -2,7 +2,6 @@ package grafana
 
 import (
 	"crypto/tls"
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"net/http/httputil"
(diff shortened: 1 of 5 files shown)


Pokom (approved):
Tested against the issue and confirmed that it can serve against sub paths.

PR #567, opened 2025-01-17
Add `--continue-on-error` support for the grr export command

### LINKED ISSUE #562: [Feature Request]: `grr export` continue on error flag
### Feature Request

The `grr push` command has a `--continue-on-error` option, which is quite useful as many of the resource files created by `grr pull` contain syntax errors and cause parse errors. The `grr export` command doesn't have such a flag, causing it to be unable to handle those invalid resources and to fail on the first one instead.

K-Phoen:
Closes #562

This makes the `grr export` function consistent with `grr apply` and `grr pull` when it comes to error management.

### DIFF
--- a/pkg/grizzly/notifier/notifier.go
+++ b/pkg/grizzly/notifier/notifier.go
@@ -2,7 +2,6 @@ package notifier
 
 import (
 	"fmt"
-	"os"
 
 	"github.com/fatih/color"
 )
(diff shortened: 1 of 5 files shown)


pkg/grizzly/notifier/notifier.go
```diff
@@ -53,15 +52,6 @@ func Info(obj fmt.Stringer, msg string) {
...
 
-// Info announces a message in green (to stderr)
-func InfoStderr(obj fmt.Stringer, msg string) {
-	if obj == nil {
-		os.Stderr.WriteString(green(msg) + "\n")
-	} else {
-		os.Stderr.WriteString(fmt.Sprintf("%s %s\n", obj.String(), green(msg)))
-	}
-}
-
```
Pokom, line 64: Was this not used?

K-Phoen: Yup, so I just drive-by deleted it :)

Pokom (approved):

Issue #555, opened 2025-01-10
[Bug]: Failed to manifest dashboard when modifying lib


wilfriedroset:
### Grizzly Version

0.6.1

### Expected Behavior

Dashboards manifestion works without error when any of the files in the watchList are modified.

### Actual Behavior

Users might want to create reusable lib. As such they will start grizzly and watch several path. This is similar to what is documented here https://grafana.github.io/grizzly/server/, this is also what is done in [grafonnet examples(https://github.com/grafana/grafonnet/tree/main/examples/redMethod). Depending on what you are work on, user could edit either the dashboard itself or the underlying reusable lib.
However when grizzly detect a change in the underlying lib the manifestion fails

```
➜  reproducer docker compose up
[+] Running 2/0
 ✔ Container grafana  Created                                                                                                                                                                            0.0s
 ✔ Container grizzly  Recreated                                                                                                                                                                          0.1s
Attaching to grafana, grizzly
grafana  | logger=plugins.registration t=2025-01-10T13:22:54.913338607Z level=error msg="Could not register plugin" pluginId=xychart error="plugin xychart is already registered"
grafana  | logger=plugins.initialization t=2025-01-10T13:22:54.913475793Z level=error msg="Could not initialize plugin" pluginId=xychart error="plugin xychart is already registered"
grizzly  | level=info msg="[watcher] Watching for changes"
grizzly  | Listening on http://localhost:8080/
grizzly  | level=error msg="[watcher] Error: parse error in '/mnt/dashboards/examples/redMethod/lib/redDashboard/main.libsonnet': RUNTIME ERROR: couldn't manifest function as JSON\n\tField \"addApplication\"\t\n\tDuring manifestation\t\n"
grizzly  | level=warning msg="[watcher] error: parse error in '/mnt/dashboards/examples/redMethod/lib/redDashboard/main.libsonnet': RUNTIME ERROR: couldn't manifest function as JSON\n\tField \"addApplication\"\t\n\tDuring manifestation\t\n"
grizzly  | level=error msg="[watcher] Error: parse error in '/mnt/dashboards/examples/redMethod/lib/redDashboard/main.libsonnet': RUNTIME ERROR: couldn't manifest function as JSON\n\tField \"addApplication\"\t\n\tDuring manifestation\t\n"
[...]

PR #575, opened 2025-01-27
Auto watch jpath

wilfriedroset:
The jpath contains additional library used to manifest the dashboards. During the development workflow one might modify the underlying lib. This commit makes grr automatically watch the jpath. Moreover, a change in the jpath will trigger the reload of all watched files.

Fixes: #555

### DIFF
--- a/cmd/grr/workflow.go
+++ b/cmd/grr/workflow.go
@@ -407,7 +407,7 @@ func serveCmd(registry grizzly.Registry) *cli.Command {
 		server.SetContext(currentContext.Name)
 		server.SetFormatting(onlySpec, format)
 		if opts.Watch {
-			server.Watch(watchPaths)
+			server.Watch(watchPaths, opts.JsonnetPaths)
 			if opts.WatchScript != "" {
 				server.WatchScript(opts.WatchScript)
 			}
--- a/pkg/grizzly/server.go
+++ b/pkg/grizzly/server.go
@@ -9,7 +9,9 @@ import (
 	"net/http/httputil"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/go-chi/chi"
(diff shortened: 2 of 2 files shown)


Pokom:
Thanks @wilfriedroset for the PR! I've pulled and tested locally and think it looks reasonable. I'm going to ask @K-Phoen to take a peak as well as they have a bit more context into this part of grizzly. 

I've allowed the PR to run relevant workflows as well, so you can get results from CI.

malcolmholmes:
A few comments. Grizzly started as a Jsonnet tool, hence Jpath being an important thing. But it evolved, with Jsonnet becoming only one way of interacting with it - no other way requires jpath.

Therefore, I'd prefer it if the "blast radius" of the jpath could be limited - if there's a watchPaths, can't the jpath dirs just be added to that right from the get go, then the rest of the code needn't change (and can stay oblivious of jpath's existence)?

wilfriedroset:
> if there's a watchPaths, can't the jpath dirs just be added to that right from the get go

This is what I initially though and tested. This is how I stumble upon the unexpected behavior I've reported here: https://github.com/grafana/grizzly/issues/555

My understanding is that the use of watchPaths to watch both the mixins (e.g: dashboards) and the lib cannot work as it is now. Indeed, when grizzly catch a change in the lib and not the mixins themselve, grizzly tries to manifest/build the lib which produce the error `RUNTIME ERROR: couldn't manifest function as JSON`

This is why I've come up with the following PR. Another idea I add while I was working on it was to add another flag, `watchPathsWithoutManifests` and any changes in those path would trigger the manifest/build of anything in `watchPaths`.

My understanding of grizzly is not as deep as I would like hence I would like your guidance @malcolmholmes about the next course of actions to fix #555

Duologic:
Like Malcom said:

> A few comments. Grizzly started as a Jsonnet tool, hence Jpath being an important thing. But it evolved, with Jsonnet becoming only one way of interacting with it - no other way requires jpath.

I'm not inclined to more broadly support jsonnet as part of Grizzly. The `grr serve --script` can do the heavy lifting instead. Please have a look at example in my comment on #555.

Duologic:
Closing this in favor of a documentation update to shell out to jsonnet instead.

PR #601, opened 2025-03-30
Add `--ignore` flag to `grr serve` command

### LINKED ISSUE #600: [Feature Request]: grr serve `--exclude` option
### Feature Request

I would like to have the `--exclude` option to specify files that are excluded target files to `grr serve` command.

We are currently managing dashboard files with the following directory structure.

```
❯ tree .
.
├── Makefile
├── ipmi-exporter
│   ├── ipmi_exporter-cluster.json
│   └── kustomization.yaml
├── kustomization.yaml
└── nvidia-gpu
├── Makefile
├── dcgm-exporter-dashboard.json
└── kustomization.yaml
```

If `grr serve --only-spec ./` is run on this directory, you will get errors for files that are not dashboard data.

Image

It would be nice if `grr serve` has  `--exclude` option and use it like this.

```
grr serve --only-spec --exclude Makefile,*.yaml ./
```

superbrothers:
This PR adds the `—ignore` flag, which specifies local files to ignore, to the `grr serve` command.

```
grr serve --ignore '{Makefile,*.yaml}' ./
```

The flag prevents errors from output when the directory contains something that is not a resource file.

Closes #600

### DIFF
--- a/cmd/grr/config.go
+++ b/cmd/grr/config.go
@@ -39,6 +39,7 @@ type Opts struct {
 	CanSave         bool
 	Watch           bool
 	WatchScript     string
+	IgnorePatterns  []string
 }
 
 func configPathCmd() *cli.Command {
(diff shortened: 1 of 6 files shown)


superbrothers:
@K-Phoen  Could you review this PR?

PR #604, opened 2025-04-06
Support executing watch script on Windows

yjoer:
It doesn't seem that Windows is officially supported, but I managed to build it with `go build`. However, I ran into an issue with the watch script while working with the Grizzly server, which is executed using `sh -c`. Since `sh` isn't available on Windows, I updated the code to perform a runtime OS check and use an equivalent shell instance `cmd /c` for Windows.

### DIFF
--- a/pkg/grizzly/server.go
+++ b/pkg/grizzly/server.go
@@ -343,7 +343,13 @@ func (s *Server) executeWatchScript() ([]byte, error) {
 	var stderr bytes.Buffer
 	log.Debugf("[watch script] executing %s", s.watchScript)
 
-	cmd := exec.Command("sh", "-c", s.watchScript)
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" {
+		cmd = exec.Command("cmd", "/c", s.watchScript)
+	} else {
+		cmd = exec.Command("sh", "-c", s.watchScript)
+	}
+
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	err := cmd.Run()


yjoer:
I'm not sure how you built it. I built it with `go build ./cmd/grr`. It's all good on my end, and I can't see any error message.

```
PS D:\grizzly> grr serve
level=info msg="Listening on http://localhost:8080/\n"
```

It will be empty because we are not pointing to any resource file.
image

yjoer:
This PR might not be needed since Grizzly is superceded by grafanactl and the differences among os are handled there [serve.go#L207-212](https://github.com/grafana/grafanactl/blob/v0.0.6/cmd/grafanactl/resources/serve.go#L207-L212)

yjoer:
I see. I coincidentally did not include `/` at the end of the URL, or I would end up in your situation. Pheww! 🤭 Thanks for telling me the impact of a slash.

yjoer:
> Although they ignored the need for windows development, your code is useful to me. I just configured the windows development environment. I will try to rebuild the binary with your code. Thank you.

They moved on to https://github.com/grafana/grafanactl. I haven't dug deep, but it looks like a continuation of this project. I can find some signs of Windows support in the release notes.

yjoer:
My workflow was about generating a JSON string from a Python script, then passing it to `grr serve`.

Changing this
```
grr serve --watch --script "uv run python script.py" .
```

Into
```
grafanactl resources serve --script "uv run python script.py" --watch .
```

Produced a lengthy parse error, saying `Object 'Kind' is missing`. I think I had to update the SDK too.

Issue #602, opened 2025-03-31
AlertRuleGroup always creating


ExzoTikStyle:
### Grizzly Version

v0.7.1

### Expected Behavior

`grr apply` should check if the `AlertRuleGroup` exists and update it

### Actual Behavior

`grr apply` always tries to create a `AlertRuleGroup`

### Steps to Reproduce

1) Get AlertRuleGroup from example
https://grafana.github.io/grizzly/grafana/#alertrulegroup
2) First run `grr apply` will create rule
3) Second run `grr apply` will failure
Logs from first run
```
grr apply . -f test -l debug
DEBUG Parsing resource                              parser=filtered resourcePath=.
DEBUG Parsing file                                  file="alert-rules/external calls.yaml" parser=yaml
Applying 1 resource
DEBUG Getting the remote value for `AlertRuleGroup.c6636739-b2ad-4c85-a026-ae3ed3e434cc.d`
DEBUG `AlertRuleGroup.c6636739-b2ad-4c85-a026-ae3ed3e434cc.d` was not found, adding it...
AlertRuleGroup.c6636739-b2ad-4c85-a026-ae3ed3e434cc.d added
1 resource added
```
Logs from second run
```
grr apply . -f udms -l debug
DEBUG Parsing resource                              parser=filtered resourcePath=.
DEBUG Parsing file                                  file="alert-rules/external calls.yaml" parser=yaml
Applying 1 resource
DEBUG Getting the remote value for `AlertRuleGroup.c6636739-b2ad-4c85-a026-ae3ed3e434cc.d`
DEBUG `AlertRuleGroup.c6636739-b2ad-4c85-a026-ae3ed3e434cc.d` was not found, adding it...
AlertRuleGroup.c6636739-b2ad-4c85-a026-ae3ed3e434cc.d failed: creating rule for group c6636739-b2ad-4c85-a026-ae3ed3e434cc.d: [POST /v1/provisioning/alert-rules][400] postAlertRuleBadRequest {"message":"a conflicting alert rule is found: rule title under the same organisation and folder should be unique"}
1 resource failed
DEBUG Silent error: 1 error occurred:
  * creating rule for group c6636739-b2ad-4c85-a026-ae3ed3e434cc.d: [POST /v1/provisioning/alert-rules][400] postAlertRuleBadRequest {"message":"a conflicting alert rule is found: rule title under the same organisation and folder should be unique"}
```

ExzoTikStyle:
[...]

PR #605, opened 2025-04-08
Fix creating AlertRuleGroup with incorrect params

ExzoTikStyle:
This PR resolve issue https://github.com/grafana/grizzly/issues/602 , when user cannot update existing `AlertRuleGroup `because of grizzly send incorrect param in ruleGroup field


gh0st3e:
Hey, thank you for this fix, i am really need this future, is this possible to merge it?

kenny-not-dead:
I agree with the above

